PR #7, opened 2016-09-06
Introducing event builder v2

aliabbasrizvi:
This diff introduces event builder v2.

This diff does not intend to:

Redirect to builder v2 based on version (will come in a separate diff)
Look for layer ID on experiment (will come in a separate diff)
Use attributes (will come in a separate diff)
Clean up terminology (change goal --> key every place et al) - Coming in a much later diff

cc @haleybash-optimizely @delikat @vraja2 @mikeng13

### DIFF
--- a/optimizely/event_builder.py
+++ b/optimizely/event_builder.py
@@ -1,4 +1,6 @@
 import time
+from abc import abstractmethod
+from abc import abstractproperty
 
 from . import version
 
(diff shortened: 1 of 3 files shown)


haleybash-optimizely:
lgtm

vraja2:
lgtm

delikat:
lgtm

PR #8, opened 2016-09-07
Picking appropriate event builder based on config version

aliabbasrizvi:
This change introduces the ability to pick appropriate event builder based on version of the datafile.

@mikeng13 @delikat @vraja2 @haleybash-optimizely


mikeng13:
LGTM

PR #18, opened 2016-09-19
Updating track event to work for grouped experiments

aliabbasrizvi:
@vraja2 @haleybash-optimizely @mikeng13 @delikat


haleybash-optimizely:
LGTM

PR #21, opened 2016-09-22
Changing to use objects

aliabbasrizvi:
Change for Experiment, Attribute and Event

@vraja2 @delikat @haleybash-optimizely @mikeng13


vraja2:
lgtm

PR #31, opened 2016-10-03
Fix parsing of v1 file

aliabbasrizvi:
@optimizely/fullstack-devs


aliabbasrizvi:
The failures are due to dependency install failures

aliabbasrizvi:
Re-ran tests and they went through fine.

vraja2:
lgtm

PR #53, opened 2017-05-20
typo - ClassName.test_name

mauerbac:


aliabbasrizvi:
Will be updating failing test as ordering is not right.

PR #63, opened 2017-06-28
Updated event_builder

mauerbac:
- Attributes list should always be included
- Removed revision number


mauerbac:
Ping @aliabbasrizvi / @delikat . See D16574

PR #88, opened 2017-10-19
Feature/event listener

thomaszurkan-optimizely:


optimizely/notification_center.py
```diff
@@ -0,0 +1,93 @@
...
+class NotificationCenter(object):
+  """ Class encapsulating Broadcast Notifications. The enums.NotifcationTypes includes predefined notifications."""
+  def __init__(self, logger):
+    self.notification_id = 1
+    self.notifications = {}
+    for attr, value in enums.NotificationTypes.__dict__.items():
+      self.notifications[value] = []
+    self.logger = logger
+
+  def add_notification(self, notification_type, notification_callback):
```
mikeng13: Can we call this `add_notification_listener` instead? It more accurately represents what the API does.

optimizely/notification_center.py
```diff
@@ -0,0 +1,93 @@
...
+          return -1
+      self.notifications[notification_type].append((self.notification_id, notification_callback))
+
+    ret_val = self.notification_id
+
+    self.notification_id += 1
+
+    return ret_val
+
+  def remove_notification(self, notification_id):
```
mikeng13: Same here: `remove_notification_listener`

optimizely/notification_center.py
```diff
@@ -0,0 +1,93 @@
...
+      notification_type: .helpers.enums.NotificationTypes
+      notification_callback: closure of function to call when event is triggered.
+    Returns:
+      notification id used to remove the notification
+    """
+    if notification_type not in self.notifications:
+      self.notifications[notification_type] = [(self.notification_id, notification_callback)]
+    else:
+      for callback in self.notifications[notification_type]:
+        if callback[1] == notification_callback:
```
mikeng13: A cleaner way of accessing the callback would be to unpack the tuple in the `for` declaration like this:

`for notification_id, callback in self.notifications[notification_type]:`

optimizely/notification_center.py
```diff
@@ -0,0 +1,93 @@
...
+  def fire_notifications(self, notification_type, *args):
+    """ Fires off the notification for the specific event.  Uses var args to pass in a
+        arbitrary list of parameter according to which notification type was fired.
+
+    Args:
+      notification_type: Type of notification to fire.
+      args: list of arguments to the callback.
+    """
+    if notification_type in self.notifications:
+      for callback in self.notifications[notification_type]:
```
mikeng13: Same as above, unpack the tuple here so you can just call `callback(*args)`

optimizely/optimizely.py
```diff
@@ -316,7 +320,8 @@ def track(self, event_key, user_id, attributes=None, event_tags=None):
       except:
         error = sys.exc_info()[1]
         self.logger.log(enums.LogLevels.ERROR, 'Unable to dispatch conversion event. Error: %s' % str(error))
-
+      self.notification_center.fire_notifications(enums.NotificationTypes.TRACK, event_key, user_id,
```
mikeng13: This actually is interesting. Do we fire the notification as soon as we dispatch the event, or we wait for the logendpoint to return to us before we tell the listener? I could see us running into counting issues (mismatches) between us and other analytics providers if we pass the event without first guaranteeing that we received it on our end. Thoughts @aliabbasrizvi @wangjoshuah @ceimaj ?

thomaszurkan-optimizely: IMHO, I think if there is a mismatch then there would be a problem on our end and it would probably bubble up faster.

mikeng13: Yeah I don't feel too strongly about it. The key here is to very clearly document in our dev docs, WHEN exactly we fire off the notification.

thomaszurkan-optimizely: well, that actually brings up a good point, do we want to have pre and post calls?

mikeng13: nah, I think that gets too complicated. The current approach you have there should be good.

optimizely/helpers/enums.py
```diff
@@ -41,3 +41,13 @@ class Errors(object):
   INVALID_VARIATION_ERROR = 'Provided variation is not in datafile.'
   UNSUPPORTED_DATAFILE_VERSION = 'Provided datafile has unsupported version. ' \
                                  'Please use SDK version 1.1.0 or earlier for datafile version 1.'
+
+
+class NotificationTypes(object):
+  """ NotificationTypes for the notification_center.NotificationCenter
+      format is NOTIFICATION TYPE: list of parameters to callback.
+  """
+  ACTIVATE = "ACTIVATE:experiment, user_id,attributes, variation, event"
```
mikeng13: What was the reason for changing this back to `ACTIVATE` from `DECISION`? `ACTIVATE` is more of an Optimizely term and might not make sense to developers of other anaytics products. Instead `DECISION` is a more industry-standard term.

mikeng13: @thomaszurkan-optimizely I left a comment here in case you miss it

optimizely/optimizely.py
```diff
@@ -383,6 +388,19 @@ def is_feature_enabled(self, feature_key, user_id, attributes=None):
...
                                     attributes)
+        self.notification_center.send_notifications(enums.NotificationTypes.FEATURE_EXPERIMENT,
+                                                  feature_key, user_id, attributes,
+                                                    decision.experiment, decision.variation)
+      else:
+        audience = None
+        if decision.experiment is not None and \
+                decision.experiment.audienceIds is not None and \
+                len(decision.experiment.audienceIds) > 0:
+          audience_id = decision.experiment.audienceIds[0]
```
oakbani: @thomaszurkan-optimizely Hi, I have a question here. In the case where the rollout rule/experiment has multiple audience IDs, and the user didn't qualify for the first audience, but for some other audience id in the list. Do we still send the audience at index 0 in send_notifications method?

PR #95, opened 2017-12-19
EventBuilder - Revisit

oakbani:
* 📝 Addressed concerns and removed interim-state from EventBuilder
* 📝  EventBuilder Logic change + Unit tests

 **eventEntityId** --  Removed //Only for checking event dispatch params. Doesn't make a difference
 **eventName** -- Removed // Same as above

**entityId/entity_id/ accountId/account_id/ projectId/project_id** -- Case issue - Both should exist in project. Data file has these in camel case while the JSON payload that we build needs to have these in snake case

 **segmentId** - Removed useless, renamed dimensions to attributes in benchmarking tests

**Note:** EventBuilder needs Bucketing ID attribute logic and it's corresponding testcase as in other SDKS. I think should be done in a separate PR to make cherry-picking easier for bucketing work

### DIFF
--- a/optimizely/event_builder.py
+++ b/optimizely/event_builder.py
@@ -67,11 +67,6 @@ class BaseEventBuilder(object):
     """
     pass
 
-  @abstractmethod
-  def _get_source(self):
-    """ Get source information. """
-    pass
-
   def _get_anonymize_ip(self):
     """ Get IP anonymization bool
 
@@ -99,7 +94,7 @@ class BaseEventBuilder(object):
       attributes: Dict representing user attributes and values which need to be recorded.
 
     Returns:
-     dict Common params required for both impression and conversion events.
+     Dict consisting of parameters common to both impression and conversion events.
     """
     commonParams = {}
 
@@ -114,8 +109,8 @@ class BaseEventBuilder(object):
     commonParams[self.EventParams.USERS].append(visitor)
     commonParams[self.EventParams.USERS][0][self.EventParams.ATTRIBUTES] = self._get_attributes(attributes)
 
-    commonParams[self.EventParams.SOURCE_SDK_TYPE] = self._get_source()[self.EventParams.SOURCE_SDK_TYPE]
-    commonParams[self.EventParams.SOURCE_SDK_VERSION] = self._get_source()[self.EventParams.SOURCE_SDK_VERSION]
+    commonParams[self.EventParams.SOURCE_SDK_TYPE] = 'python-sdk'
+    commonParams[self.EventParams.SOURCE_SDK_VERSION] = version.__version__
     commonParams[self.EventParams.ANONYMIZE_IP] = self._get_anonymize_ip()
 
     return commonParams
@@ -158,7 +153,7 @@ class EventBuilder(BaseEventBuilder):
       attributes: Dict representing user attributes and values which need to be recorded.
 
     Returns:
-      list Valid attributes for the user (if any). Otherwise, empty.
+      List consisting of valid attributes for the user. Empty otherwise.
     """
 
     params = []
@@ -181,18 +176,6 @@ class EventBuilder(BaseEventBuilder):
 
     return params
 
-  def _get_source(self):
-    """ Get source information.
-
-    Returns:
-      dict SDK type and version information.
-    """
-    params = {}
-    params[self.EventParams.SOURCE_SDK_TYPE] = 'python-sdk'
-    params[self.EventParams.SOURCE_SDK_VERSION] = version.__version__
-
-    return params
-
   def _get_required_params_for_impression(self, experiment, variation_id):
     """ Get parameters that are required for the impression event to register.
 
@@ -201,7 +184,7 @@ class EventBuilder(BaseEventBuilder):
       variation_id: ID for variation which would be presented to user.
 
     Returns:
-      dict Decisions and events info for impression event.
+      Dict consisting of decisions and events info for impression event.
     """
     snapshot = {}
 
@@ -229,7 +212,7 @@ class EventBuilder(BaseEventBuilder):
       decisions: List of tuples representing valid experiments IDs and variation IDs.
 
     Returns:
-      dict Decisions and events info for conversion event.
+      Dict consisting of the decisions and events info for conversion event.
     """
 
     for experiment_id, variation_id in decisions:
--- a/optimizely/event_builder.py
+++ b/optimizely/event_builder.py
@@ -1,4 +1,4 @@
-# Copyright 2016-2017, Optimizely
+# Copyright 2016-2018, Optimizely
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
@@ -44,7 +44,7 @@ class BaseEventBuilder(object):
     """ Get project ID.
 
     Returns:
-      Project ID in the datafile.
+      Project ID of the datafile.
     """
 
     return self.config.get_project_id()
@@ -57,15 +57,6 @@ class BaseEventBuilder(object):
     """
 
     return self.config.get_account_id()
-
-  @abstractmethod
-  def _get_user(self, user_id):
-    """ Get user.
-
-    Args:
-      user_id: ID of the user.
-    """
-    pass
 
   @abstractmethod
   def _get_attributes(self, attributes):
@@ -115,8 +106,12 @@ class BaseEventBuilder(object):
     commonParams[self.EventParams.PROJECT_ID] = self._get_project_id()
     commonParams[self.EventParams.ACCOUNT_ID] = self._get_account_id()
 
+    visitor = {}
+    visitor[self.EventParams.END_USER_ID] = user_id
+    visitor[self.EventParams.SNAPSHOTS] = []
+
     commonParams[self.EventParams.USERS] = []
-    commonParams[self.EventParams.USERS].append(self._get_user(user_id))
+    commonParams[self.EventParams.USERS].append(visitor)
     commonParams[self.EventParams.USERS][0][self.EventParams.ATTRIBUTES] = self._get_attributes(attributes)
 
     commonParams[self.EventParams.SOURCE_SDK_TYPE] = self._get_source()[self.EventParams.SOURCE_SDK_TYPE]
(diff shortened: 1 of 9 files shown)


optimizely/event_builder.py
```diff
@@ -139,54 +145,57 @@ class EventParams(object):
...
     if not attributes:
-      return
+      return []
 
     for attribute_key in attributes.keys():
       attribute_value = attributes.get(attribute_key)
-      # Omit falsy attribute values
-      if attribute_value:
+      # Do not discard if value is zero or false
+      if attribute_value is not None:
```
oakbani: attribute value can be integer zero or boolean false.

tests/test_event_dispatcher.py
```diff
@@ -44,10 +44,10 @@ def test_dispatch_event__post_request(self):
 
     url = 'https://www.optimizely.com'
     params = {
-      'accountId': '111001',
-      'eventName': 'test_event',
-      'eventEntityId': '111028',
-      'visitorId': 'oeutest_user'
+      'a': '111001',
```
oakbani: These testcases only check that event is dispatched with exactly the same params. Actual params do not make a difference.

msohailhussain:
@mikeng13 @aliabbasrizvi 
no reviewer is assigned yet, can u please assign any reviewer or review it.

tests/test_event_dispatcher.py
aliabbasrizvi: I fail to understand the need to do this. Why are we changing things for the sake of changing things? This is beyond the scope of the change and contributes to the review size.

optimizely/event_builder.py
```diff
@@ -267,12 +274,13 @@ def create_impression_event(self, experiment, variation_id, user_id, attributes)
       Event object encapsulating the impression event.
     """
 
-    self.params = {}
-    self._add_common_params(user_id, attributes)
-    self._add_required_params_for_impression(experiment, variation_id)
+    params = self._get_common_params(user_id, attributes)
```
aliabbasrizvi: Please restore to the previous state. There is no need for getters here.

optimizely/event_builder.py
aliabbasrizvi: Attributes are supposed to be strings always. So the previous check was valid.

optimizely/event_builder.py
```diff
@@ -35,78 +35,84 @@ class BaseEventBuilder(object):
...
 
     Returns:
       Current time in milliseconds.
     """
 
     return int(round(time.time() * 1000))
 
-  def _add_common_params(self, user_id, attributes):
-    """ Add params which are used same in both conversion and impression events.
+  def _get_common_params(self, user_id, attributes):
```
aliabbasrizvi: Please restore to the previous state.

tests/benchmarking/data.py
```diff
@@ -281,58 +281,57 @@
   "audiences": [
     {
       "conditions": "[\"and\", [\"or\", [\"or\", {\"name\": \"browser_type\", "
-                    "\"type\": \"custom_dimension\", \"value\": \"safari\"}]]]",
```
aliabbasrizvi: These are correct as this benchmarking data is from our legacy product. Please restore all this.

optimizely/event_builder.py
msohailhussain: according to documentation, it can accept string, boolean and int. right now, no test case when attributes value should be  like {"is_local": false} which would be discarded by old code but actually it shouldn't be discarded.

tests/test_event_dispatcher.py
msohailhussain: it was giving some misunderstanding like we are using eventName, eventEntityId, just for the sake of removing any misconception, we just renamed it.

optimizely/event_builder.py
msohailhussain: this change was needed because existing code is not thread safe & all other SDKs using the same approach.

msohailhussain:
@aliabbasrizvi Please let me know if we need to discuss anything relevant to this PR. This PR is more relevant to refactoring and adopting changes like in other SDKs.

optimizely/event_builder.py
aliabbasrizvi: @msohailhussain it should be discarded and the existing behavior is correct.

tests/test_event_dispatcher.py
aliabbasrizvi: Please undo the renaming.

optimizely/event_builder.py
aliabbasrizvi: Can you elaborate on the thread safety aspect?

aliabbasrizvi:
I'd recommend keeping this diff small and not changing anything unless absolutely necessary. Other enhancements can be done separately.

optimizely/event_builder.py
msohailhussain: in the old code, when create_conversion_event is called, it reset params property
https://github.com/optimizely/python-sdk/blob/master/optimizely/event_builder.py#L293

and if createconversion event is called in parallel twice using same optimizely object, the later one will overwrite first params values because param is global in the EventBuilder class. and EventBuilder is initialized only once in the Optimizely class.
https://github.com/optimizely/python-sdk/blob/master/optimizely/optimizely.py#L82

optimizely/event_builder.py
msohailhussain: ok sure, we will revert it.

tests/test_event_dispatcher.py
msohailhussain: ok, we will revert it.

msohailhussain:
@aliabbasrizvi Please let me know if thread-safe stuff we have to keep or revert. I discussed with Mike as well and he wanted this change in python as well. The bug was more related to eventbuilder refactoring, that's why we made this change.

optimizely/event_builder.py
aliabbasrizvi: Ok. Then keep the change and update the docstrings to reflect what is being returned in each method which you changed.

aliabbasrizvi:
@msohailhussain you can keep the thread safe stuff, but update docstrings in all getters to reflect what they are returning. You can look up examples of what the docstring should look like by seeing the `Returns` section of different methods in the repo.

aliabbasrizvi:
Thanks for making all these changes.

optimizely/event_builder.py
```diff
@@ -35,78 +35,95 @@ class BaseEventBuilder(object):
...
     pass
 
-  def _add_project_id(self):
-    """ Add project ID to the event. """
+  def _get_project_id(self):
+    """ Get project ID.
 
-    self.params[self.EventParams.PROJECT_ID] = self.config.get_project_id()
+    Returns:
+      Project ID in the datafile.
```
aliabbasrizvi: `.. of the datafile`

optimizely/event_builder.py
```diff
@@ -35,78 +35,95 @@ class BaseEventBuilder(object):
...
+      Account ID in the datafile.
+    """
 
-    self.params[self.EventParams.ACCOUNT_ID] = self.config.get_account_id()
+    return self.config.get_account_id()
 
   @abstractmethod
-  def _add_user_id(self, user_id):
-    """ Add user ID to the event.
+  def _get_user(self, user_id):
```
aliabbasrizvi: This does not make sense as you are passing it `user_id` which is already available

tests/test_event_builder.py
```diff
@@ -174,8 +255,7 @@ def test_create_conversion_event__with_attributes(self):
     }
 
     with mock.patch('time.time', return_value=42.123), \
```
aliabbasrizvi: Why was this test changed?

tests/test_event_builder.py
```diff
@@ -139,6 +139,87 @@ def test_create_impression_event__with_attributes(self):
...
+          self.project_config.get_experiment_from_key('test_experiment'),
+          '111129', 'test_user', {'do_you_know_me': 'test_value'}
+        )
+      self._validate_event_object(event_obj,
+                                  event_builder.EventBuilder.EVENTS_URL,
+                                  expected_params,
+                                  event_builder.EventBuilder.HTTP_VERB,
+                                  event_builder.EventBuilder.HTTP_HEADERS)
+
+  def test_create_conversion_event(self):
```
aliabbasrizvi: Why are these tests being added?

oakbani: I found these two scenarios missing, which we have covered in other SDKs. 
1. test that expected impression payload is created when an attribute, which is not in the datafile, is given. 
2. test that expected conversion event is created when no attributes are passed at all.

tests/test_event_builder.py
oakbani: We don't need to mock bucketer when testing these EventBuilder methods.

optimizely/event_builder.py
oakbani: Removed this method. And brought logic out into get_common_params method.

optimizely/event_builder.py
```diff
@@ -35,78 +35,90 @@ class BaseEventBuilder(object):
...
-    """ Add params which are used same in both conversion and impression events.
+  def _get_common_params(self, user_id, attributes):
+    """ Get params which are used same in both conversion and impression events.
 
     Args:
       user_id: ID for user.
       attributes: Dict representing user attributes and values which need to be recorded.
+
+    Returns:
+     dict Common params required for both impression and conversion events.
```
aliabbasrizvi: Change this to `Dict consisting of parameters common to both impression and conversion events.`

optimizely/event_builder.py
```diff
@@ -139,58 +151,57 @@ class EventParams(object):
...
-  def _add_attributes(self, attributes):
-    """ Add attribute(s) information to the event.
+  def _get_attributes(self, attributes):
+    """ Get attribute(s) information.
 
     Args:
       attributes: Dict representing user attributes and values which need to be recorded.
+
+    Returns:
+      list Valid attributes for the user (if any). Otherwise, empty.
```
aliabbasrizvi: This commenting style is not common in Python. Please change this to 

`List consisting of valid attributes for the user. Empty otherwise.`

optimizely/event_builder.py
```diff
@@ -139,58 +151,57 @@ class EventParams(object):
...
+  def _get_source(self):
+    """ Get source information.
 
-  def _add_user_id(self, user_id):
-    """ Add user ID to the event.
-
-    Args:
-      user_id: ID of the user.
+    Returns:
+      dict SDK type and version information.
```
aliabbasrizvi: No need to set this like this. Rather just set it like the user ID is set i.e. in place.

tests/test_event_builder.py
aliabbasrizvi: We do need to as it confirms that the test will show the right variation.

aliabbasrizvi: Please restore the mock and then this will be good to go.

msohailhussain:
plz. wait am fixing it.

msohailhussain:
@aliabbasrizvi have added bucketer mocking

aliabbasrizvi:
Merged in https://github.com/optimizely/python-sdk/pull/99

PR #102, opened 2018-02-16
Feature Toggle

oakbani:


optimizely/entities.py
```diff
@@ -109,7 +109,8 @@ def __init__(self, id, value, **kwards):
       self.id = id
       self.value = value
 
-  def __init__(self, id, key, variables=None, **kwargs):
+  def __init__(self, id, key, featureEnabled=False, variables=None, **kwargs):
```
mikeng13: Actually, just realized this might not work. Only feature experiments will have this property. Regular experiment variations will omit this property and you will run into a situation where the variation comes in with just `id, key, and variables` so you'd need to accommodate that scenario

oakbani: Shouldn't be an issue. generate_key_map in project_config calls as key,value arguments. 
Still, have added a new test case to test the scenario.

tests/test_config.py
```diff
@@ -973,6 +985,25 @@ def test_init__with_v4_datafile(self):
     self.assertEqual(expected_rollout_id_map, project_config.rollout_id_map)
     self.assertEqual(expected_variation_variable_usage_map, project_config.variation_variable_usage_map)
 
+  def test_variation_has_featureEnabled_false_if_prop_undefined(self):
+    """ Test that featureEnabled property by default is set to False, when not given in the data file"""
+    variation = {
```
mikeng13: Please indent this correctly

tests/test_config.py
```diff
@@ -973,6 +985,25 @@ def test_init__with_v4_datafile(self):
...
+            }, {
+              'id': '129',
+              'value': '112'
+            }, {
+              'id': '130',
+              'value': '1.211'
+            }]}
+
+    variation_entity = entities.Variation(**variation)
+    self.assertFalse(variation_entity.featureEnabled)
```
mikeng13: Can you also assert that the rest of the properties are properly populated?

tests/test_config.py
oakbani: Do you mean 4 space indent? The whole file right now is on a 2 space indent.

mikeng13: Hmm github just shows it weird. See how line 1003 (the closing bracket) does not match the indentation of the opening bracket on line 990

mikeng13:
build

mikeng13:
build

PR #106, opened 2018-03-19
parses revenue value

rashidsp:


rashidsp:
Closing it. Need to update logger in event-tag-utils

PR #113, opened 2018-04-05
Fix feature variable accessor default value

oakbani:
Please note that we check for **'featureEnabled'** flag only in the method **'isFeatureEnabled'**. So this flag makes no difference in any of the variable accessor methods. 

The issue at hand was that variations, which have an empty **'variables'** node, didn't get included in **'variation_variable_usage_map'**. This resulted in https://github.com/optimizely/python-sdk/blob/master/optimizely/project_config.py#L435 case when detailed logs were inspected. 

**Changes in place:** 
* Generate a key for every variation in **'variation_variable_usage_map'**
* Move log message and add additional message to provide correct and verbose logs.

### DIFF
--- a/tests/test_optimizely.py
+++ b/tests/test_optimizely.py
@@ -1436,7 +1436,7 @@ class OptimizelyTest(base.BaseTest):
 
     mock_logger.assert_called_once_with(
       enums.LogLevels.INFO,
-      'Value for variable "is_working" of feature flag "test_feature_in_experiment" is true for user "test_user".'
+      'Value for variable "is_working" for variation "variation" is "true".'
     )
 
   def test_get_feature_variable_double(self):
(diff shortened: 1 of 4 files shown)


tests/test_optimizely.py
```diff
@@ -1436,7 +1436,7 @@ def test_get_feature_variable_boolean(self):
 
     mock_logger.assert_called_once_with(
       enums.LogLevels.INFO,
-      'Value for variable "is_working" of feature flag "test_feature_in_experiment" is true for user "test_user".'
+      'Value for variable "is_working" for variation "variation" is "true".'
```
oakbani: * feature_flag removed as we don't have feature key reference in the new method where we have moved this log message. Additionally, in logging mode logs mentioning feature keys precede this log so no need of it. 
* user_id same, not available. And the message sort of implied that selection of value somehow depends on the user ID which isn't true.

PR #114, opened 2018-04-09
Improve code coverage

oakbani:


optimizely/project_config.py
```diff
@@ -575,9 +575,6 @@ def get_forced_variation(self, experiment_key, user_id):
       return None
 
     variation = self.get_variation_from_id(experiment_key, variation_id)
-    if not variation:
```
oakbani: This can be safely removed since we check and test that a valid variation is set in the map in set_forced_variation.

mikeng13:
build

PR #121, opened 2018-05-31
Introduce Bot Filtering

oakbani:


optimizely/event_builder.py
```diff
@@ -172,19 +182,39 @@ def _get_attributes(self, attributes):
     if not attributes:
       return []
 
-    for attribute_key in attributes.keys():
+    for attribute_key in sorted(attributes.keys()):
```
msohailhussain: why do we need to sort?

optimizely/event_builder.py
```diff
@@ -172,19 +182,39 @@ def _get_attributes(self, attributes):
...
       attribute_value = attributes.get(attribute_key)
       # Omit falsy attribute values
       if attribute_value:
-        attribute = self.config.get_attribute(attribute_key)
-        if attribute:
+        # Check for reserved attributes
+        reserved_attrs = [ReservedAttributes.BUCKETING_ID, ReservedAttributes.USER_AGENT]
```
msohailhussain: this must be outside, as it will be initialized again and again.

optimizely/event_builder.py
```diff
@@ -172,19 +182,39 @@ def _get_attributes(self, attributes):
...
+            params.append({
+              'entity_id': attribute.id,
+              'key': attribute_key,
+              'type': self.EventParams.CUSTOM,
+              'value': attribute_value,
+            })
+
+    # Append Bot Filtering Attribute
+    attribute_key = ReservedAttributes.BOT_FILTERING
+    params.append({
```
msohailhussain: Please add condition which was placed in javascript code. If no botFiltering is given or old datafile is used, in that case it shouldn't send the value of botFiltering.
https://github.com/optimizely/javascript-sdk/pull/99/files#diff-1fbd9d0f02956a420aeef6959fc24caaR73

optimizely/project_config.py
```diff
@@ -56,6 +56,7 @@ def __init__(self, datafile, logger, error_handler):
     self.feature_flags = config.get('featureFlags', [])
     self.rollouts = config.get('rollouts', [])
     self.anonymize_ip = config.get('anonymizeIP', False)
+    self.bot_filtering = config.get('botFiltering', False)
```
msohailhussain: default is NULL, not false. Please recheck it.

optimizely/event_builder.py
```diff
@@ -172,19 +182,39 @@ def _get_attributes(self, attributes):
...
+            'entity_id': attribute_key,
             'key': attribute_key,
             'type': self.EventParams.CUSTOM,
-            'value': attribute_value,
+            'value': attribute_value
           })
 
+        else:
+          attribute = self.config.get_attribute(attribute_key)
+          if attribute:
```
msohailhussain: in case attribute is not found, pruning that attribute but where's the log? Add log or I would suggest revise it as per javascript code.

tests/test_config.py
```diff
@@ -679,6 +681,20 @@ def test_get_project_id(self):
 
     self.assertEqual(self.config_dict['projectId'], self.project_config.get_project_id())
 
+  def test_get_bot_filtering(self):
+    """ Test that bot filtering is retrieved correctly when using get_bot_filtering_value. """
+
+    # Assert bot filtering is False when not provided in data file
+    self.assertTrue('botFiltering' not in self.config_dict)
+    self.assertEqual(False, self.project_config.get_bot_filtering_value())
```
msohailhussain: I need to check, when botFiltering is not in dictionary what it's supposed to return? it MUST be NULL, i haven't seen such condition in JS.

tests/base.py
```diff
@@ -150,6 +150,7 @@ def setUp(self):
       'accountId': '12001',
       'projectId': '111111',
       'version': '4',
+      'botFiltering': False,
```
msohailhussain: It may be included or not included, don't add it here. Let's discuss tonight.

msohailhussain:
First revise the code and align with JS & then correct test cases. Right now, not reviewing test cases anymore.

optimizely/event_builder.py
oakbani: attributes is a dict. Order of keys is not guaranteed to be same for all python versions. Hence for eventpayload unit tests to pass for all versions, we need to ensure that they get appended in the same order.

optimizely/event_builder.py
```diff
@@ -18,6 +18,7 @@
 
 from . import version
 from .helpers import event_tag_utils
+from .helpers.enums import ControlAttributes
```
aliabbasrizvi: nit. I would usually just import enums and refer to `enums.ControlAttributes`

optimizely/event_builder.py
```diff
@@ -85,6 +86,15 @@ def _get_anonymize_ip(self):
 
     return self.config.get_anonymize_ip_value()
 
+  def _get_bot_filtering(self):
+    """ Get bot filtering bool
+
+    Returns:
+      'botFiltering' value in the datafile.
```
aliabbasrizvi: nit. `Boolean representing whether bot filtering is enabled or not.`

I would similarly also change the comment for IP anonymization in the method above to 

Boolean representing whether IP anonymization is enabled or not.

optimizely/event_builder.py
aliabbasrizvi: Why does ordering matter? Let us not do this. It seems unnecessary.

optimizely/event_builder.py
```diff
@@ -172,19 +182,28 @@ def _get_attributes(self, attributes):
...
           params.append({
-            self.EventParams.EVENT_ID: attribute.id,
+            'entity_id': attribute_id,
             'key': attribute_key,
             'type': self.EventParams.CUSTOM,
             'value': attribute_value,
           })
 
+    # Append Bot Filtering Attribute
+    if isinstance(self._get_bot_filtering(), bool):
```
aliabbasrizvi: Nice.

optimizely/event_builder.py
```diff
@@ -172,19 +182,28 @@ def _get_attributes(self, attributes):
...
+      params.append({
+          'entity_id': ControlAttributes.BOT_FILTERING,
+          'key': ControlAttributes.BOT_FILTERING,
+          'type': self.EventParams.CUSTOM,
+          'value': self._get_bot_filtering(),
```
aliabbasrizvi: Lets call `self._get_bot_filtering` at line 198 and use that.

optimizely/event_builder.py
oakbani: @aliabbasrizvi  attributes is a dict. Order of keys is not guaranteed to be same for all python versions. Hence for eventpayload unit tests to pass for all versions, we need to ensure that they get appended in the same order.
https://travis-ci.org/optimizely/python-sdk/builds/389076889

optimizely/event_builder.py
```diff
@@ -169,21 +179,29 @@ def _get_attributes(self, attributes):
...
-        attribute = self.config.get_attribute(attribute_key)
-        if attribute:
-          params.append({
-            self.EventParams.EVENT_ID: attribute.id,
-            'key': attribute_key,
-            'type': self.EventParams.CUSTOM,
-            'value': attribute_value,
-          })
+    if isinstance(attributes, dict):
+      for attribute_key in sorted(attributes.keys()):
```
aliabbasrizvi: Do no sort these.

optimizely/project_config.py
```diff
@@ -363,20 +366,28 @@ def get_event(self, event_key):
...
 
     if attribute:
-      return attribute
+      if has_reserved_prefix:
+        self.logger.warning(('Attribute %s unexpectedly has reserved prefix %s; using attribute ID '
+                             'instead of reserved attribute name.' % (attribute_key, RESERVED_ATTRIBUTE_PREFIX)))
+
+      return attribute.id
+
+    if has_reserved_prefix and attribute_key != enums.ControlAttributes.BOT_FILTERING:
```
aliabbasrizvi: Why the not condition? I do not thing it is needed. Lets remove the second condition.

tests/test_config.py
```diff
@@ -679,6 +681,20 @@ def test_get_project_id(self):
...
+    """ Test that bot filtering is retrieved correctly when using get_bot_filtering_value. """
+
+    # Assert bot filtering is None when not provided in data file
+    self.assertTrue('botFiltering' not in self.config_dict)
+    self.assertIsNone(self.project_config.get_bot_filtering_value())
+
+    # Assert bot filtering is retrieved as provided in the data file
+    opt_obj = optimizely.Optimizely(json.dumps(self.config_dict_with_features))
+    project_config = opt_obj.config
+    self.assertEqual(self.config_dict_with_features['botFiltering'],
```
aliabbasrizvi: nit. This styling seems wrong
```
self.assertEqual(
  self.config_dict_with_features['botFiltering'],
  project_config.get_bot_filtering_value()
)
```

tests/test_config.py
```diff
@@ -787,16 +803,32 @@ def test_get_event__invalid_key(self):
...
+  def test_get_attribute_id__reserved_key(self):
+    """ Test that Attribute Key is returned as ID when provided attribute key is reserved key. """
+    self.assertEqual('$opt_user_agent',
+                       self.project_config.get_attribute_id('$opt_user_agent'))
+
+  def test_get_attribute_id__unknown_key_with_opt_prefix(self):
+    """ Test that Attribute Key is returned as ID when provided attribute key is not
+    present in the datafile but has $opt prefix. """
+    self.assertEqual('$opt_interesting',
+                       self.project_config.get_attribute_id('$opt_interesting'))
```
aliabbasrizvi: nit. Indentation seems off. s of self should align with the '

oakbani:
@aliabbasrizvi Please review the changes.

PR #124, opened 2018-06-04
Updating description and adding long description

aliabbasrizvi:


setup.py
```diff
@@ -20,7 +20,11 @@
 setup(
     name='optimizely-sdk',
     version=__version__,
-    description="SDK for Optimizely's Full Stack Python project.",
+    description='Python SDK for Optimizely X Full Stack.',
+    long_description='Optimizely X Full Stack is A/B testing and feature management for product development teams. '
```
amilstead: A more standard approach to this is to set the value of `long_description` to the contents of a `README.md` (or similar) prepended to the contents of the `CHANGELOG.md` (or similar).

e.g.

```
with open('README.md') as _file:
  README = _file.read()
with open('CHANGELOG.md') as _file:
  CHANGELOG = _file.read()

setup(
  ...
  long_description=README + '\n\n# Change Log: \n' + CHANGELOG
  ...
)
```

Any thoughts?

aliabbasrizvi: Yeah i noticed that in some implementations.

How about we do the message mentioned here + README + CHANGELOG?

amilstead: SGTM

amilstead: Don't forget to add those files in the `MANIFEST.in` so they're included in the package distribution. Installs will fail if without them.

aliabbasrizvi: Good call. Need README there.

setup.py
```diff
@@ -17,10 +17,22 @@
...
+with open(os.path.join(here, 'README.md')) as _file:
+  README = _file.read()
+
+with open(os.path.join(here, 'CHANGELOG.md')) as _file:
+  CHANGELOG = _file.read()
+
+about_text = 'Optimizely X Full Stack is A/B testing and feature management for product development teams. ' \
+             'Experiment in any application. Make every feature on your roadmap an opportunity to learn. ' \
+             'Learn more at https://www.optimizely.com/products/full-stack/ or see our documentation at ' \
+             'https://developers.optimizely.com/x/solutions/sdks/reference/index.html?language=python'
```
spencerwilson-optimizely: nit: maybe a period at the end?

aliabbasrizvi: Good call. Will add.

PR #135, opened 2018-08-21
Jongsy/fix multiple experiment event params

jongsy:

PR #136, opened 2018-08-21
fix(track): Send decisions for all experiments using an event when using track

aliabbasrizvi:


aliabbasrizvi:
This includes commits from #135

optimizely/event_builder.py
```diff
@@ -243,40 +243,39 @@ def _get_required_params_for_conversion(self, event_key, event_tags, decisions):
...
+      numeric_value = event_tag_utils.get_numeric_value(event_tags, self.config.logger)
+      if numeric_value is not None:
+        event_dict[event_tag_utils.NUMERIC_METRIC_TYPE] = numeric_value
 
-        snapshot[self.EventParams.EVENTS] = [event_dict]
+      if len(event_tags) > 0:
+        event_dict[self.EventParams.TAGS] = event_tags
 
-        return snapshot
+    snapshot.setdefault(self.EventParams.EVENTS, []).append(event_dict)
```
mikeng13: Do we need to setdefault as opposed to just initializing an empty array with that `event_dict`?

tests/base.py
```diff
@@ -375,5 +375,152 @@ def setUp(self):
       }]
     }
 
-    self.optimizely = optimizely.Optimizely(json.dumps(self.config_dict))
+    self.config_dict_with_multiple_experiments = {
```
mikeng13: Shouldn't this be passed in for that one unit test that uses it?

aliabbasrizvi: I think it is ok

optimizely/event_builder.py
aliabbasrizvi: Yeah, let me update that.

aliabbasrizvi:
build-e2e

Issue #139, opened 2018-08-29
Built in EventDispatch silently fails


allanlei:
The default `EventDispatcher` only handles network related errors and ignores HTTP errors(`500`, `503`, etc) without any logging messages

```
try:
  if event.http_verb == enums.HTTPVerbs.GET:
    requests.get(event.url, params=event.params, timeout=10)
  elif event.http_verb == enums.HTTPVerbs.POST:
    requests.post(event.url, json=event.params, headers=event.headers, timeout=10)
except request_exception.RequestException as error:
  logging.error('Dispatch event failed. Error: %s' % str(error))
```

HTTP errors in `requests` do not raise Exceptions
```
In [2]: requests.get('https://httpbin.org/status/500')
Out[2]: <Response [500]>   # This is an object, not an raise error therefore not caught by try/except
```

You would need to raise from the response
```
response = requests.get(event.url, params=event.params, timeout=10)
In [4]: response.raise_for_status()
---------------------------------------------------------------------------
HTTPError                                 Traceback (most recent call last)
<ipython-input-4-46d4bf99ee16> in <module>()
----> 1 requests.get('https://httpbin.org/status/500').raise_for_status()

python3.6/site-packages/requests/models.py in raise_for_status(self)
    937 
    938         if http_error_msg:
--> 939             raise HTTPError(http_error_msg, response=self)
    940 
    941     def close(self):

HTTPError: 500 Server Error: INTERNAL SERVER ERROR for url: https://httpbin.org/status/500
```

aliabbasrizvi:
Hi @allanlei. Thank you for logging this issue. Will take a look at this.

aliabbasrizvi:
@allanlei this issue was fixed in https://github.com/optimizely/python-sdk/pull/140 and will be available in next release of the SDK. 

Closing this issue and will comment when next release is available.

PR #140, opened 2018-08-29
fix(dispatch): Updating default event dispatcher to actually raise exceptions

aliabbasrizvi:
Address issue raised [here](https://github.com/optimizely/python-sdk/issues/139)

### DIFF
--- a/optimizely/event_dispatcher.py
+++ b/optimizely/event_dispatcher.py
@@ -34,8 +34,11 @@ class EventDispatcher(object):
 
     try:
       if event.http_verb == enums.HTTPVerbs.GET:
-        requests.get(event.url, params=event.params, timeout=REQUEST_TIMEOUT)
+        requests.get(event.url, params=event.params, timeout=REQUEST_TIMEOUT).raise_for_status()
       elif event.http_verb == enums.HTTPVerbs.POST:
-        requests.post(event.url, data=json.dumps(event.params), headers=event.headers, timeout=REQUEST_TIMEOUT)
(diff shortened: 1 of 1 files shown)


optimizely/event_dispatcher.py
```diff
@@ -34,8 +34,11 @@ def dispatch_event(event):
...
       if event.http_verb == enums.HTTPVerbs.GET:
-        requests.get(event.url, params=event.params, timeout=REQUEST_TIMEOUT)
+        requests.get(event.url, params=event.params, timeout=REQUEST_TIMEOUT).raise_for_status()
       elif event.http_verb == enums.HTTPVerbs.POST:
-        requests.post(event.url, data=json.dumps(event.params), headers=event.headers, timeout=REQUEST_TIMEOUT)
+        requests.post(
+          event.url, data=json.dumps(event.params), headers=event.headers, timeout=REQUEST_TIMEOUT
+        ).raise_for_status()
+
     except request_exception.RequestException as error:
```
nchilada: It looks like this would catch any `requests.HTTPError`, which makes me wonder whether the PR is titled correctly. At the very least, I would describe it in the release notes as "Updated default event dispatcher to log HTTP errors".

@aliabbasrizvi what do you think?

nchilada: And by the way, should we consider calling the customer's error handler as well?

Or should we _actually_ raise upon certain kinds of errors? Instead of catching and logging from within the default event dispatcher.

PR #142, opened 2018-09-05
 feat(api) : Accepting all types of Attribute values

oakbani:
## Summary
1. Allow all attribute value types to be used for targeting purposes.
2. Filter out attribute values that are not supported from the event payload.


tests/test_optimizely.py
```diff
@@ -544,6 +544,80 @@ def test_activate__with_attributes__audience_match(self):
...
+        'optimizely.bucketer.Bucketer.bucket',
+        return_value=self.project_config.get_variation_from_id('test_experiment', '111129')) \
+        as mock_bucket, \
+      mock.patch('time.time', return_value=42), \
+      mock.patch('uuid.uuid4', return_value='a68cf1ad-0393-4e18-af87-efe8f01a7c9c'), \
+      mock.patch('optimizely.event_dispatcher.EventDispatcher.dispatch_event') as mock_dispatch_event:
+
+      attributes = {
+          'test_attribute': 'test_value_1',
+          'boolean_attribute': True,
```
oakbani: @mikeng13 This fails when I set it to False, because of the condition here: https://github.com/optimizely/python-sdk/blob/master/optimizely/event_builder.py#L186
This condition removes all attributes with falsy values including False, 0 and 0.0

Should I fix it in this PR?

oakbani:
Closing it for now. Needs some work.

optimizely/decision_service.py
```diff
@@ -34,20 +35,28 @@ def __init__(self, config, user_profile_service):
...
+  def _get_bucketing_id(self, user_id, attributes):
     """ Helper method to determine bucketing ID for the user.
 
     Args:
       user_id: ID for user.
       attributes: Dict representing user attributes. May consist of bucketing ID to be used.
 
     Returns:
-      String representing bucketing ID for the user. Fallback to user's ID if not provided.
+      String representing bucketing Id if it is a String type in attributes else return userId.
```
aliabbasrizvi: nit. Please update this to be user ID and bucketing ID instead of variable names

optimizely/decision_service.py
```diff
@@ -34,20 +35,28 @@ def __init__(self, config, user_profile_service):
...
     """
 
     attributes = attributes or {}
-    return attributes.get(enums.ControlAttributes.BUCKETING_ID, user_id)
+    bucketing_id = attributes.get(enums.ControlAttributes.BUCKETING_ID)
+
+    if bucketing_id is not None:
+      if isinstance(bucketing_id, string_types):
+        self.logger.debug('Bucketing ID is valid: "%s".' % (bucketing_id))
```
aliabbasrizvi: I think this message is not useful as it will happen often that bucketing ID is fine and it is a lot of new unnecessary messageds. I'd drop this and just use the warning message.

optimizely/helpers/validator.py
```diff
@@ -167,3 +167,28 @@ def is_non_empty_string(input_id_key):
...
+    True otherwise
+  """
+
+  if not isinstance(attribute_key, string_types):
+    return False
+
+  if isinstance(attribute_value, string_types):
+    return True
+
+  if type(attribute_value) in (int, float, bool):
```
aliabbasrizvi: Just combine the 2 ifs together through OR

tests/helpers_tests/test_condition.py
```diff
@@ -33,10 +33,29 @@ def setUp(self):
     self.condition_evaluator = condition_helper.ConditionEvaluator(self.condition_list, attributes)
 
   def test_evaluator__returns_true(self):
-    """ Test that evaluator correctly returns True when there is a match. """
+    """ Test that evaluator correctly returns True when there is an exact match.
+    Also test that evaluator works for falsy values. """
 
+    # string attribute value
+    condition_list = [['test_attribute', ""]]
```
aliabbasrizvi: nit. Single quotes for empty string.

tests/helpers_tests/test_validator.py
```diff
@@ -146,6 +146,25 @@ def test_is_non_empty_string(self):
     self.assertTrue(validator.is_non_empty_string('0'))
     self.assertTrue(validator.is_non_empty_string('test_user'))
 
+  def test_is_attribute_valid(self):
+    """ Test that non-string attribute key or unsupported attribute value returns False."""
+
```
aliabbasrizvi: nit. Just annotate that here you are testing for invalid attribute keys by putting a comment. Similarly for the next 2 sections.

optimizely/helpers/validator.py
```diff
@@ -167,3 +167,25 @@ def is_non_empty_string(input_id_key):
     return True
 
   return False
+
+
+def is_attribute_valid(attribute_key, attribute_value):
+  """ Determine if given attribute is valid.
```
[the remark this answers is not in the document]
msohailhussain: @mikeng13 both keys & strings are validated here. @nchilada Discussed with Nikhail, in dynamic languages key can be int/bool and won't raise compile time issue. So we need to check attribute key as well. i guess 'is_attribute_valid' is OK.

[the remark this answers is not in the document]
nchilada: Yup, I'd actually leave this up to SDK implementation - we could explicitly filter out _or_ we could keep attributes with non-string keys, as long as we never we coerce the keys from non-strings to strings. A datafile's audiences will only target using string keys.

optimizely/decision_service.py
```diff
@@ -34,20 +35,27 @@ def __init__(self, config, user_profile_service):
...
+  def _get_bucketing_id(self, user_id, attributes):
     """ Helper method to determine bucketing ID for the user.
 
     Args:
       user_id: ID for user.
       attributes: Dict representing user attributes. May consist of bucketing ID to be used.
 
     Returns:
-      String representing bucketing ID for the user. Fallback to user's ID if not provided.
+      String representing bucketing ID if it is a String type in attributes else return user ID.
```
oakbani: @mikeng13 Do we also accept empty string as bucketing ID?

mikeng13: Empty string is a valid user ID right now so it makes sense to allow it for bucketing ID as well.

optimizely/optimizely.py
```diff
@@ -212,6 +212,9 @@ def _get_feature_variable_for_type(self, feature_key, variable_key, variable_typ
       self.logger.error(enums.Errors.NONE_USER_ID_PARAMETER)
       return None
 
+    if not self._validate_user_inputs(attributes):
```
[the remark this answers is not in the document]
msohailhussain: Just want to summarize, only null and maps are allowed which is already covered in the code.

PR #145, opened 2018-10-05
refact(ConditionEvaluator): Simplify condition evaluation

oakbani:
Python SDK as of now, has a bit complex condition evaluation. Before modifying evaluator for audience match type work, this simplifies the condition evaluator and makes it more like how it's done in Ruby, PHP.

Currently, In python we have conditions list and structure.
The structure stores decoded conditions string with { } replaced by an integer placeholder.
And the conditions list is a list of those replaced { } conditions.
This PR removes this undesired complexity. And directly decodes conditions string as in other SDKs.
This will make audience type work easier, where we will be checking match type/ attribute types. etc.

### DIFF
--- a/optimizely/entities.py
+++ b/optimizely/entities.py
@@ -27,11 +27,10 @@ class Attribute(BaseEntity):
 
 class Audience(BaseEntity):
 
-  def __init__(self, id, name, conditions, conditionStructure=None, conditionList=None, **kwargs):
+  def __init__(self, id, name, conditions, conditionList=None, **kwargs):
     self.id = id
     self.name = name
     self.conditions = conditions
-    self.conditionStructure = conditionStructure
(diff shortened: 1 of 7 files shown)


aliabbasrizvi:
Can you elaborate as to why this change is needed at the moment?

My only hesitance is that this has been around for a while and not worth changing in light of the changes we will be making in the near term. This same structure/list paradigm will be useful when we do boolean audiences.

msohailhussain:
@aliabbasrizvi It's very true, @thomaszurkan-optimizely had the same comment in a discussion for boolean audiences for this approach. But to align with other SDKs, we wanted to make this change. I can discuss in-person tomorrow. @mikeng13

aliabbasrizvi:
Going to close this as previously mentioned.

PR #146, opened 2018-10-25
feat (audience match types): Update condition evaluator for new audience match types

oakbani:
### Summary

This adds support for new audience match type conditions to the condition evaluator:
- exact, exists, gt, lt, and substring conditions
- Abort and return null when appropriate in leaf evaluators
- Null handling in and/or/not evaluators
- Assume implicit "or" when operator is missing

### Test Plan

Added new unit tests.
Ran compatibility suite tests for typed audiences


tests/base.py
```diff
@@ -19,6 +19,12 @@
 
 class BaseTest(unittest.TestCase):
 
+  def assertStrictTrue(self, to_assert):
+    self.assertIs(to_assert, True)
+
+  def assertStrictFalse(self, to_assert):
+    self.assertIs(to_assert, False)
+
```
oakbani: assertTrue and assertFalse check for truthy and falsy values. 
Added these to ensure that we don't mistake null for false

.github/pull_request_template.rst
```diff
@@ -0,0 +1,15 @@
+Summary
```
msohailhussain: why this changes are coming in this PR? @oakbani

oakbani: Updated base branch with master

optimizely/helpers/condition.py
```diff
@@ -12,115 +12,348 @@
...
     """
+    saw_null_result = False
 
     for condition in conditions:
-      result = self.evaluate(condition)
+      result = self.evaluate(condition, leaf_evaluator)
       if result is False:
         return False
+      if result is None:
+        saw_null_result = True
```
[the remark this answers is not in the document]
oakbani: So that we return FALSE, if there is one.

aliabbasrizvi:
I think compat tests are now available. You can update the test plan.

optimizely/helpers/condition.py
```diff
@@ -12,115 +12,348 @@
...
-      if conditions[0] in DEFAULT_OPERATOR_TYPES:
-        return self.OPERATORS[conditions[0]](self, conditions[1:])
-      else:
-        return False
+    if self.condition_data[index][2] != self.CUSTOM_ATTRIBUTE_CONDITION_TYPE:
+      return None
+
+    condition_match = self.condition_data[index][3]
+
+    if condition_match not in self.MATCH_TYPES:
```
nchilada: Nit: here, too, I wonder if we really need `MATCH_TYPES`. Something like
```python
if condition_match not in self.EVALUATORS_BY_MATCH_TYPE:
```
seems more natural given that we're using this to gate the expression
```
self.EVALUATORS_BY_MATCH_TYPE[condition_match]
```

optimizely/helpers/condition.py
```diff
@@ -12,115 +12,348 @@
...
+    if not isinstance(value, (numbers.Integral, float)):
+      # numbers.Integral instead of int to accomodate long integer in python 2
+      return False
+
+    if isinstance(value, bool):
+      # bool is a subclass of int
       return False
 
-    return not self.evaluate(single_condition[0])
+    if math.isnan(value) or math.isinf(value):
```
nchilada: Nit: the `isinstance(value, (numbers.Integral, float))` check at the top of this function should help, but I think it might be safer to only call `isnan`/`isinf` when `isinstance(value, float)`.
```python
if isinstance(value, float):
  if math.isnan(value) or math.isinf(value):
    return False
```
This would protect us if `isnan`/`isinf` were to start throwing `TypeError` when given integers in some future version of Python. It's [not](https://docs.python.org/3.7/library/math.html#math.isinf) [documented](https://docs.python.org/3.7/library/math.html#math.isinf), but these functions already throw a `TypeError` when given non-numeric values (as determined in my Python 2.7.10 and 3.7.0 REPLs).

optimizely/helpers/condition.py
```diff
@@ -12,115 +12,348 @@
...
+
     """
-    if len(single_condition) != 1:
+    if not len(conditions) > 0:
+      return None
+
+    result = self.evaluate(conditions[0], leaf_evaluator)
+    return None if result is None else not result
+
+  DEFAULT_OPERATOR_TYPES = [
```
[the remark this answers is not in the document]
oakbani: I was wondering if we should create **condition** as a separate module and keep each of these classes in separate files for better organization. @nchilada  Yeah we can make them static. 
@aliabbasrizvi  What do you suggest? The changes will propagate to project config and combinations work(which has been done). I hope we are not in an urgency to merge this

optimizely/helpers/audience.py
```diff
@@ -24,8 +24,16 @@ def is_match(audience, attributes):
...
-  condition_evaluator = condition_helper.ConditionEvaluator(audience.conditionList, attributes)
-  return condition_evaluator.evaluate(audience.conditionStructure)
+  condition_tree_evaluator = condition_helper.ConditionTreeEvaluator()
+  custom_attr_condition_evaluator = condition_helper.CustomAttributeConditionEvaluator(
+    audience.conditionList, attributes)
+
+  is_match = condition_tree_evaluator.evaluate(
+    audience.conditionStructure,
+    lambda index: custom_attr_condition_evaluator.evaluate(index)
+  )
```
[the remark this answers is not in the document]
oakbani: Though requires a lot of rework, I am still willing to get #145 merged. This will make it easier for later work especially audience combinations.

optimizely/helpers/condition.py
```diff
@@ -12,115 +12,348 @@
...
+    """ Method to validate if the given value is a number and not one of NAN, INF, -INF
+
+    Args:
+      value: Value to be validated
+
+    Returns:
+      Boolean: True if value is a number and not NAN, INF or -INF else False
+    """
+    if not isinstance(value, (numbers.Integral, float)):
+      # numbers.Integral instead of int to accomodate long integer in python 2
```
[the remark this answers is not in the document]
oakbani: Yeah, is_attribute_valid should be updated in that case. 
Also note that helpers.validator.is_attribute_valid is only called when creating event payload. 
For audience evaluation and listener called, we use all of the user passed attributes

optimizely/helpers/condition.py
```diff
@@ -12,115 +12,348 @@
...
+    first_val_type = type(first_val)
+    second_val_type = type(second_val)
+
+    # use isinstance to accomodate Python 2 unicode and str types
+    if isinstance(first_val, string_types) and isinstance(second_val, string_types):
+      return True
+
+    # Compare types if one of the values is bool because bool is a subclass on Integer
+    if isinstance(first_val, bool) or isinstance(second_val, bool):
+      return first_val_type == second_val_type
```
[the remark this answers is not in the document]
oakbani: In that case if only 1 of values is a bool type and the other is float or int
it will return true in the following if block as bool is a subclass of int

optimizely/helpers/condition.py
```diff
@@ -159,7 +392,12 @@ def _audience_condition_deserializer(obj_dict):
   Returns:
     List consisting of condition key and corresponding value.
   """
-  return [obj_dict.get('name'), obj_dict.get('value')]
+  return [
+    obj_dict.get('name'),
+    obj_dict.get('value'),
+    obj_dict.get('type'),
+    obj_dict.get('match', ConditionMatchTypes.EXACT)
```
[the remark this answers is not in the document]
oakbani: @aliabbasrizvi

optimizely/helpers/condition.py
```diff
@@ -12,115 +12,348 @@
...
+    """
+    condition_value = self.condition_data[index][1]
+    user_value = self.attributes.get(self.condition_data[index][0])
+
+    if not isinstance(condition_value, string_types) or not isinstance(user_value, string_types):
+      return None
+
+    return condition_value in user_value
+
+  EVALUATORS_BY_MATCH_TYPE = {
```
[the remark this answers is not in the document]
oakbani: I suppose can't declare it before defining evaluator methods

optimizely/helpers/condition_tree_evaluator.py
```diff
@@ -0,0 +1,124 @@
...
+    leaf_evaluator: Function which will be called to evaluate leaf condition values
+
+  Returns:
+    Boolean: Result of evaluating the conditions using the operator rules and the leaf evaluator.
+    None: if conditions couldn't be evaluated
+
+  """
+
+  if isinstance(conditions, list):
+    if conditions[0] in OPERATOR_TYPES:
```
oakbani: @nchilada Can't use EVALUATORS_BY_OPERATOR_TYPE here as in implicit OR case the first item in conditions[0] is a dict which is an unhashable type. Using EVALUATORS_BY_OPERATOR_TYPE.items() has different result types for python 2 and 3

optimizely/helpers/validator.py
```diff
@@ -189,3 +191,57 @@ def is_attribute_valid(attribute_key, attribute_value):
...
+def is_finite_number(value):
+  """ Method to validate if the given value is a number and not one of NAN, INF, -INF
+
+  Args:
+    value: Value to be validated
+
+  Returns:
+    Boolean: True if value is a number and not NAN, INF or -INF else False
+  """
+  if not isinstance(value, (int, float)):
```
nchilada: `numbers.Integral` instead of `int`

optimizely/helpers/validator.py
```diff
@@ -189,3 +191,57 @@ def is_attribute_valid(attribute_key, attribute_value):
...
+
+  if isinstance(value, bool):
+    # bool is a subclass of int
+    return False
+
+  if isinstance(value, float):
+    if math.isnan(value) or math.isinf(value):
+      return False
+
+  return True
```
nchilada: Thanks for extracting this! As mentioned on a now-obsolete commit, can we

1. enforce an absolute size limit of 1e53 if `isinstance(value, numbers.Integral)`
2. update `helpers.validator.is_attribute_valid` such that it calls `helpers.validator.is_finite_number`

?

optimizely/helpers/condition.py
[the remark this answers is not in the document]
nchilada: Oh wow, you're right. So tricky! I haven't looked at the unit tests yet but hopefully we have test cases for this. The compatibility suite can't test the specifics of Python data types

optimizely/helpers/condition.py
nchilada: @aliabbasrizvi I think that question is better suited for the conversation(s) in which we're talking about calling `is_finite` from `is_attribute_valid`

At any rate, it looks like @oakbani accepted the original suggestion in this convo, so I'm going to resolve 😬

optimizely/helpers/validator.py
nchilada: And @aliabbasrizvi's addendum:

> Random. I am wondering if we should extend this validation i.e. the `is_finite` check to revenue and quantity values?

+1, is worth considering

optimizely/helpers/condition.py
```diff
@@ -12,115 +12,348 @@
...
+      # numbers.Integral instead of int to accomodate long integer in python 2
+      return False
+
+    if isinstance(value, bool):
+      # bool is a subclass of int
       return False
 
-    return not self.evaluate(single_condition[0])
+    if math.isnan(value) or math.isinf(value):
+      return False
```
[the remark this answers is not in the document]
nchilada: I'm resolving this conversation but have ported the comment to the new location of this function (in `helpers.validator`.

optimizely/helpers/condition.py
```diff
@@ -12,115 +12,348 @@
...
     Args:
       conditions: List of conditions ex: [operand_1, operand_2]
+      leaf_evaluator: Function which will be called to evaluate leaf condition values
 
     Returns:
-      Boolean: True if all operands evaluate to True
+      Boolean:
+        - True if all operands evaluate to True
+        - False if a single operand evaluates to False
+      None: if conditions couldn't be evaluated
```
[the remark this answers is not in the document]
nchilada: Not sure. Would definitely get verbose. Perhaps we could just force ourselves to be diligent in documenting the new, collective behavior somewhere else? E.g. in the KB or at the very top of the condition tree module (rather than separately for each operator's evaluator)

optimizely/helpers/condition.py
```diff
@@ -12,115 +12,348 @@
...
+      Returns:
+        Boolean:
+          - True if the user attribute value is greater than the condition value
+          - False if the user attribute value is less than or equal to the condition value
+        None: if the condition value isn't finite or the user attribute value isn't finite
+    """
+    condition_value = self.condition_data[index][1]
+    user_value = self.attributes.get(self.condition_data[index][0])
+
+    if not self.is_finite(condition_value) or not self.is_finite(user_value):
```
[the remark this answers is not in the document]
nchilada: Yes, `not self.is_finite(condition_value)` is degenerate (though still worth checking).

We're refining our log messages as a separate initiative...it's looking like we're more concerned about logging problems that customers are *actually* expected to encounter, but we'll see. Can I add you as a reviewer on the other doc that I'm talking about?

optimizely/helpers/condition.py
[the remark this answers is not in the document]
nchilada: Hmm, okay, thanks @oakbani.

You're right, at least for the retval format - it's easier to just wait and refactor everything together! #145

For the `'exact'` default, there is a slight risk that we'll forget to change this in the future, causing us to accidentally use the wrong default for some future condition type. So for that one I'd still prefer to make the change now, if not too distracting from the bigger things in this PR

optimizely/helpers/audience.py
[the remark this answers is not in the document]
nchilada: Interesting. Thanks for offering to take that on! I'm going to resolve this conversation, will discuss elsewhere once this PR is stable/merged

optimizely/helpers/condition_tree_evaluator.py
nchilada: > Using EVALUATORS_BY_OPERATOR_TYPE.keys() has different result types for python 2 and 3

I'm actually not very familiar with Python 3, but could we try
```python
if conditions[0] in EVALUATORS_BY_OPERATOR_TYPE:
```
? Since that's the equivalent of
```python
if conditions[0] in EVALUATORS_BY_OPERATOR_TYPE.keys():
```
in Python 2

optimizely/helpers/validator.py
oakbani: Will do this in a separate commit.

nchilada: I assume by "separate commit" you mean "separate PR"? That's fine, thanks! Tracking in 3654

tests/helpers_tests/test_condition.py
```diff
@@ -17,117 +17,342 @@
...
 
+browserConditionSafari = ['browser_type', 'safari', 'custom_attribute', 'exact']
+booleanCondition = ['is_firefox', True, 'custom_attribute', 'exact']
+integerCondition = ['num_users', 10, 'custom_attribute', 'exact']
+doubleCondition = ['pi_value', 3.14, 'custom_attribute', 'exact']
 
-class ConditionEvaluatorTests(base.BaseTest):
+exists_condition_list = [['input_value', None, 'custom_attribute', 'exists']]
+exact_string_condition_list = [['favorite_constellation', 'Lacerta', 'custom_attribute', 'exact']]
+exact_number_condition_list = [['lasers_count', 9000, 'custom_attribute', 'exact']]
```
nchilada: Can we also have a test for exact number match where the condition value is a float? I see the `doubleCondition` above, but it looks like that's used for something else

tests/helpers_tests/test_condition.py
```diff
@@ -17,117 +17,342 @@
...
+    evaluator = condition_helper.CustomAttributeConditionEvaluator(
+      exact_string_condition_list, {}
+    )
+
+    self.assertIsNone(evaluator.evaluate(0))
+
+  def test_exact_number__returns_true__when_user_provided_value_is_equal_to_condition_value(self):
+
+    evaluator = condition_helper.CustomAttributeConditionEvaluator(
+      exact_number_condition_list, {'lasers_count': 9000}
```
nchilada: Can we also find a way to test with `9000L` in the case of Python 2?

nchilada: Actually...*do* we run our unit tests in both Python 2 and Python 3?? @aliabbasrizvi

tests/helpers_tests/test_condition.py
```diff
@@ -17,117 +17,342 @@
...
+
+  def test_exact_string__returns_null__when_no_user_provided_value(self):
+
+    evaluator = condition_helper.CustomAttributeConditionEvaluator(
+      exact_string_condition_list, {}
+    )
+
+    self.assertIsNone(evaluator.evaluate(0))
+
+  def test_exact_number__returns_true__when_user_provided_value_is_equal_to_condition_value(self):
```
nchilada: We should have two versions of this:
* `test_exact_int_returns_true__when_user_provided_value_is_equal_to_condition_value`
* `test_exact_float_returns_true__when_user_provided_value_is_equal_to_condition_value`
with the same assertions but different condition value types.

Same for most or all of the other `exact_number`, `greater_than`, and `less_than` tests!

tests/helpers_tests/test_condition.py
```diff
@@ -17,117 +17,342 @@
...
+doubleCondition = ['pi_value', 3.14, 'custom_attribute', 'exact']
 
-class ConditionEvaluatorTests(base.BaseTest):
+exists_condition_list = [['input_value', None, 'custom_attribute', 'exists']]
+exact_string_condition_list = [['favorite_constellation', 'Lacerta', 'custom_attribute', 'exact']]
+exact_number_condition_list = [['lasers_count', 9000, 'custom_attribute', 'exact']]
+exact_bool_condition_list = [['did_register_user', False, 'custom_attribute', 'exact']]
+substring_condition_list = [['headline_text', 'buy now', 'custom_attribute', 'substring']]
+gt_condition_list = [['meters_travelled', 48.2, 'custom_attribute', 'gt']]
+lt_condition_list = [['meters_travelled', 48.2, 'custom_attribute', 'lt']]
```
nchilada: Can we also have `int` versions of the above two conditions?

tests/helpers_tests/test_condition.py
```diff
@@ -12,122 +12,587 @@
...
 import mock
+from six import PY2, PY3
 
 from optimizely.helpers import condition as condition_helper
 
 from tests import base
 
+if PY3:
+  def long(a):
+    pass
```
nchilada:
```suggestion
    raise NotImplementedError('Tests should only call `long` if running in PY2')
```

nchilada: (I'm guessing you only need to define `long` because the interpreter can't determine that the calls to `long` are unreachable in PY3?)

oakbani: Yeah. Exactly

optimizely/helpers/condition.py
```diff
@@ -12,115 +12,180 @@
 # limitations under the License.
 
 import json
+import numbers
 
+from six import string_types
 
-class ConditionalOperatorTypes(object):
+from .validator import is_finite_number, are_values_same_type
```
aliabbasrizvi: nit. I personally prefer doing something like:

`from . import validator` and then call the respective method.

optimizely/helpers/condition.py
```diff
@@ -12,115 +12,180 @@
...
+    self.attributes = attributes or {}
 
-  def evaluator(self, condition):
-    """ Method to compare single audience condition against provided user data i.e. attributes.
+  def is_value_valid_for_exact_conditions(self, value):
+    """ Method to validate if the value is valid for exact match type evaluation
 
     Args:
-      condition: Integer representing the index of condition_data that needs to be used for comparison.
+      value: Value to validate
```
aliabbasrizvi: nit. Period (.) at the end of line. Please make sure to do so in all messages in this PR.

optimizely/helpers/condition.py
```diff
@@ -12,115 +12,180 @@
...
 
     Returns:
-      Boolean indicating the result of comparing the condition value against the user attributes.
+      Boolean: True if value is a string type, or a boolean, or is finite. Otherwise False
```
aliabbasrizvi: nit. Same as above.

optimizely/helpers/condition.py
```diff
@@ -12,115 +12,180 @@
...
 
-  def and_evaluator(self, conditions):
-    """ Evaluates a list of conditions as if the evaluator had been applied
-    to each entry and the results AND-ed together
+  def exact_evaluator(self, index):
+    """ Evaluate the given exact match condition for the user attributes
 
     Args:
-      conditions: List of conditions ex: [operand_1, operand_2]
+      index: Index of the condition to be evaluated
```
aliabbasrizvi: index is a built in Python method I think. Please rename to something else.

optimizely/helpers/condition.py
```diff
@@ -12,115 +12,180 @@
...
+       not are_values_same_type(condition_value, user_value):
+        return None
+
+    return condition_value == user_value
+
+  def exists_evaluator(self, index):
+    """ Evaluate the given exists match condition for the user attributes
+
+      Args:
+        index: Index of the condition to be evaluated
```
aliabbasrizvi: Same comment as above. Please update all instances of index in this PR to be something different.

tests/test_config.py
```diff
@@ -120,13 +120,13 @@ def test_init(self):
         '11154', 'Test attribute users 1',
         '["and", ["or", ["or", {"name": "test_attribute", "type": "custom_attribute", "value": "test_value_1"}]]]',
         conditionStructure=['and', ['or', ['or', 0]]],
-        conditionList=[['test_attribute', 'test_value_1']]
+        conditionList=[['test_attribute', 'test_value_1', 'custom_attribute', None]]
```
aliabbasrizvi: We haven't updated `project_config.py` code. How is this getting set to None?

optimizely/helpers/condition.py
nchilada: Objects have a special method named `__index__`, but no global `index` variable, I don't think. See the global functions reference in [2](https://docs.python.org/2/library/functions.html) and [3](https://docs.python.org/3/library/functions.html).

optimizely/helpers/condition.py
nchilada: I think this is actually clean after all!

tests/test_config.py
oakbani: We have edited `def _audience_condition_deserializer(obj_dict):` to add type and match to attribute node so that we use it in `CustomAttributeConditionEvaluator - evaluate()`
https://github.com/optimizely/python-sdk/pull/146/files#diff-54739e7fb9eb3054932a2f39734e8676R218-R232

PR #147, opened 2018-10-26
feat (audience match types): Update audience evaluator and project config for new audience match types

oakbani:
### Summary

This updates project config and the audience evaluator to finish the implementation of typed audience evaluation:

- In audience_evaluator, allow evaluation to continue when no user attributes are passed (previously we were immediately returning `false` when no attributes were passed)
- Update project_config to update audience_id_map with typed_audiences
- Added unit tests checking that all top-level methods that accept user attributes can bucket users into experiments or rollouts that use typed audiences.

### Test Plan

- New & existing unit tests
- Ran compatibility suite tests for typed audiences

### DIFF
--- a/optimizely/helpers/condition.py
+++ b/optimizely/helpers/condition.py
@@ -15,7 +15,7 @@ import json
 import math
 import numbers
 
-from six import string_types, PY2
+from six import string_types
 
 
 class ConditionOperatorTypes(object):
@@ -197,6 +197,35 @@ class CustomAttributeConditionEvaluator(object):
 
     return False
 
+  def are_values_same_type(self, first_val, second_val):
+    """ Method to verify that both values belong to same type. Float and integer are
+    considered as same type.
+
+    Args:
+      first_val: Value to validate
+      second_Val: Value to validate
+
+    Returns:
+      Boolean: True if both values belong to same type. Otherwise False
+    """
+
+    first_val_type = type(first_val)
+    second_val_type = type(second_val)
+
+    # use isinstance to accomodate Python 2 unicode and str types
+    if isinstance(first_val, string_types) and isinstance(second_val, string_types):
+      return True
+
+    # Compare types if one of the values is bool because bool is a subclass on Integer
+    if isinstance(first_val, bool) or isinstance(second_val, bool):
+      return first_val_type == second_val_type
+
+    # Treat ints and floats as same type
+    if isinstance(first_val, (numbers.Integral, float)) and isinstance(second_val, (numbers.Integral, float)):
+        return True
+
+    return False
+
   def exact_evaluator(self, index):
     """ Evaluate the given exact match condition for the user attributes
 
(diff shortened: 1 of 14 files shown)


optimizely/helpers/condition.py
```diff
@@ -174,6 +174,10 @@ def is_finite(self, value):
       # numbers.Integral instead of int to accomodate long integer in python 2
       return False
 
+    if isinstance(value, bool):
```
aliabbasrizvi: Doesn't this belong in https://github.com/optimizely/python-sdk/pull/146?

tests/base.py
```diff
@@ -595,6 +595,254 @@ def setUp(self, config_dict='config_dict'):
...
+          'id': '3468206643',
+          'name': '$$dummyExactBoolean',
+          'conditions': '{ "type": "custom_attribute", "name": "$opt_dummy_attribute", "value": "impossible_value" }'
+        }
+      ],
+      'typedAudiences': [
+        {
+          'id': '3988293898',
+          'name': 'substringString',
+          'conditions': '["and", ["or", ["or", '
```
aliabbasrizvi: After recent discussions, `conditions` will not be stringified but instead object (or list of objects and boolean operators)

aliabbasrizvi: @nchilada ^

oakbani: I see. Let me revise it.

optimizely/helpers/condition.py
oakbani: Yeah, but this showed up only when I as able to run new audiences compat suite tests with this PR. Will move it into #146.

tests/test_config.py
```diff
@@ -764,6 +764,43 @@ def test_get_audience__invalid_id(self):
 
     self.assertIsNone(self.project_config.get_audience('42'))
 
+  def test_get_audience__retrieves_audiences_first_from_typedAudiences_before_audiences(self):
```
nchilada: Nit: can we rename this to something like `test_get_audience__prefers_typedAudiences_over_audiences`? To me, the current name suggests that the SDK **does** eventually look at `audiences` even if something was found in `typedAudiences`

tests/test_optimizely.py
```diff
@@ -2046,6 +2046,134 @@ def test_get_feature_variable__returns_none_if_unable_to_cast(self):
 
     mock_client_logger.error.assert_called_with('Unable to cast value. Returning None.')
```
nchilada: It looks like our test cases are (mostly) grouped by API. Can we collate the new tests appropriately?

tests/helpers_tests/test_condition.py
```diff
@@ -588,7 +588,7 @@ def test_less_than__returns_false__when_user_value_not_less_than_condition_value
   def test_less_than__returns_null__when_user_value_is_not_a_number(self):
 
     evaluator = condition_helper.CustomAttributeConditionEvaluator(
-      self.lt_condition_list, {'meters_travelled': 'a long way'}
+      self.lt_condition_list, {'meters_travelled': False}
```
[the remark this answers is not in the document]
oakbani: Yeah. Will move this into #146 .  Came up when running compat suite. Boolean is a sub class of int in python. Added the boolean check in code and modified this test to better check the failing scenario.

tests/test_config.py
```diff
@@ -764,6 +764,43 @@ def test_get_audience__invalid_id(self):
...
+    self.assertTrue(audience_3988293898 in audiences)
+
+    typed_audience_3988293898 = {
+      'id': '3988293898',
+      'name': 'substringString',
+      'conditions': ['and', ['or', ['or', {'name': 'house', 'type': 'custom_attribute',
+                     'match': 'substring', 'value': 'Slytherin'}]]]
+    }
+
+    self.assertTrue(typed_audience_3988293898 in typed_audiences)
```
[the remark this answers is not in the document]
oakbani: The actual test case assertion is below this. 
The above two are just to show that these audiences **do** exist in the datafile.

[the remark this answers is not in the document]
nchilada: Ahh, I see, you're checking `self.config_dict_with_typed_audiences['audiences']`/`self.config_dict_with_typed_audiences['typedAudiences']` rather than `opt.config.audiences`/`opt.config.typed_audiences`. Seems fine, though it might be nice to comment that this is just a sanity check for our tests

PR #149, opened 2018-11-13
fix(set_forced_variation): Treats empty variation key as invalid and does not reset forced variation.

rashidsp:
Summary
-------

- Forced variation key passed as an empty String will be invalid argument.
- Updated inputs validation for set_forced_variation.
- Updated unit tests for empty string variation key.

### DIFF
--- a/tests/test_config.py
+++ b/tests/test_config.py
@@ -1055,8 +1055,11 @@ class ConfigTest(base.BaseTest):
 
     self.assertFalse(self.project_config.set_forced_variation('test_experiment', 'test_user',
                                                               'variation_not_in_datafile'))
-    self.assertTrue(self.project_config.set_forced_variation('test_experiment', 'test_user', ''))
     self.assertTrue(self.project_config.set_forced_variation('test_experiment', 'test_user', None))
+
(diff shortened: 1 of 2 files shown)


tests/test_config.py
```diff
@@ -1055,9 +1055,12 @@ def test_set_forced_variation__invalid_variation_key(self):
 
     self.assertFalse(self.project_config.set_forced_variation('test_experiment', 'test_user',
                                                               'variation_not_in_datafile'))
-    self.assertTrue(self.project_config.set_forced_variation('test_experiment', 'test_user', ''))
     self.assertTrue(self.project_config.set_forced_variation('test_experiment', 'test_user', None))
 
+    with mock.patch.object(self.project_config, 'logger') as mock_config_logging:
+      self.assertIs(self.project_config.set_forced_variation('test_experiment', 'test_user', ''), False)
```
aliabbasrizvi: nit. Use `self.assertFalse` instead.

oakbani: assertFalse asserts Falsy value which includes None. assert_is ensures that we are returning strictly False from the method.

PR #150, opened 2018-12-04
feat (audiences): Audience combinations

oakbani:
Summary
-------
This adds support for audience combinations on experiments. If `experiment.audienceConditions` is present, it will be used as a condition tree where the leaf conditions are audience Ids.

- The audience evaluator uses the condition tree evaluator on both the experiment audience conditions and the individual audience conditions.
- The audience evaluator treats conditions trees and flat lists of audience ids (the old format) the same way, because a flat list of audience ids is a valid condition tree with an implicit OR condition.
- Added a new method to Experiment entity that returns audienceConditions if not None, otherwise returns audience Ids

Test plan
---------
- test_optimizely - Added unit tests for APIs that pass/fail using various complex audiences with different match types. 
- test_audience - Added new unit tests for calling condition tree and custom attribute evaluators. 
- Ran complex audiences test cases in compatibility test suite. 

Issues
------


optimizely/helpers/audience.py
```diff
@@ -50,17 +29,34 @@ def is_user_in_experiment(config, experiment, attributes):
...
+    audience = config.get_audience(audienceId)
+    custom_attr_condition_evaluator = condition_helper.CustomAttributeConditionEvaluator(
+      audience.conditionList, attributes)
+
+    return custom_attr_condition_evaluator.evaluate(index)
 
-    if is_match(audience, attributes):
-      return True
+  def evaluate_audience(audienceId):
+    audience = config.get_audience(audienceId)
```
nchilada: What if the audience can't be found? I think we should gracefully return null instead of throwing an AttributeError when we dereference `audience.conditionStructure`.

My apologies - I didn't think about this until now, so it wasn't in the spec 😬 I'll add a compatibility test for this! It looks like Java and JS are already compliant (although they differ in whether they invoke the error handler; I'd say we shouldn't).

tests/helpers_tests/test_audience.py
```diff
@@ -11,98 +11,174 @@
...
+    experiment = self.project_config.get_experiment_from_key('test_experiment')
+    with mock.patch('optimizely.helpers.condition_tree_evaluator.evaluate', return_value=None) as cond_tree_eval:
+
+      self.assertStrictFalse(audience.is_user_in_experiment(self.project_config, experiment, user_attributes))
+
+    with mock.patch('optimizely.helpers.condition_tree_evaluator.evaluate', return_value=False) as cond_tree_eval:
+
+      self.assertStrictFalse(audience.is_user_in_experiment(self.project_config, experiment, user_attributes))
+
+  def test_is_user_in_experiment__evaluates_audience_Ids(self):
```
nchilada: Nit: maybe delete the underscore in `audience_Ids`?

PR #151, opened 2018-12-10
feat(attribute_value): Don't target NAN, INF, -INF and  > 2^53

oakbani:
Summary
-------
Don't target NAN, INF or -INF doubles
Don't target Values > 2^53. 
Modified is_attribute_valid to check if the number is finite. 

Test plan
---------

Issues
------
OASIS-3654
https://github.com/optimizely/python-sdk/pull/146#discussion_r237850806

### DIFF
--- a/optimizely/helpers/validator.py
+++ b/optimizely/helpers/validator.py
@@ -13,6 +13,8 @@
 
 import json
 import jsonschema
+import math
+import numbers
 from six import string_types
 
 from optimizely.user_profile import UserProfile
@@ -185,7 +187,40 @@ def is_attribute_valid(attribute_key, attribute_value):
   if not isinstance(attribute_key, string_types):
     return False
 
-  if isinstance(attribute_value, string_types) or type(attribute_value) in (int, float, bool):
+  if isinstance(attribute_value, (string_types, bool)):
     return True
 
+  if isinstance(attribute_value, (numbers.Integral, float)):
+    return is_finite_number(attribute_value)
+
   return False
+
+
+def is_finite_number(value):
+  """ Validates if the given value is a number, enforces
+  limit of 1e53 for integers and restricts NAN, INF, -INF for doubles.
+
+  Args:
+    value: Value to be validated
+
+  Returns:
+    Boolean: True if value is a finite number else False
+  """
+
+  if not isinstance(value, (numbers.Integral, float)):
+      # numbers.Integral instead of int to accomodate long integer in python 2
+    return False
+
+  if isinstance(value, bool):
+    # bool is a subclass of int
+    return False
+
+  if isinstance(value, numbers.Integral):
+    if value > 1e53:
+      return False
+
+  if isinstance(value, float):
+    if math.isnan(value) or math.isinf(value):
+      return False
+
+  return True
--- a/optimizely/helpers/validator.py
+++ b/optimizely/helpers/validator.py
@@ -204,7 +204,8 @@ def is_finite_number(value):
     value: Value to be validated.
 
   Returns:
-    Boolean: True if value is a number and not NAN, INF, -INF or greater than 2^53 else False.
+    Boolean: True if value is a number and not NAN, INF, -INF or
+             greater than absolute limit of 2^53 else False.
   """
   if not isinstance(value, (numbers.Integral, float)):
       # numbers.Integral instead of int to accomodate long integer in python 2
--- a/optimizely/helpers/validator.py
+++ b/optimizely/helpers/validator.py
@@ -187,20 +187,24 @@ def is_attribute_valid(attribute_key, attribute_value):
   if not isinstance(attribute_key, string_types):
     return False
 
-  if isinstance(attribute_value, string_types) or type(attribute_value) in (int, float, bool):
+  if isinstance(attribute_value, (string_types, bool)):
     return True
+
+  if isinstance(attribute_value, (numbers.Integral, float)):
+    return is_finite_number(attribute_value)
 
   return False
 
 
 def is_finite_number(value):
-  """ Method to validate if the given value is a number and not one of NAN, INF, -INF.
+  """ Validates if the given value is a number, enforces
+   absolute limit of 2^53 and restricts NAN, INF, -INF.
 
   Args:
     value: Value to be validated.
 
   Returns:
-    Boolean: True if value is a number and not NAN, INF or -INF else False.
+    Boolean: True if value is a number and not NAN, INF, -INF or greater than 2^53 else False.
   """
   if not isinstance(value, (numbers.Integral, float)):
       # numbers.Integral instead of int to accomodate long integer in python 2
@@ -213,6 +217,9 @@ def is_finite_number(value):
   if isinstance(value, float):
     if math.isnan(value) or math.isinf(value):
       return False
+
+  if abs(value) > (2**53):
+    return False
 
   return True
 
--- a/tests/helpers_tests/test_condition.py
+++ b/tests/helpers_tests/test_condition.py
@@ -283,25 +283,26 @@ class CustomAttributeConditionEvaluator(base.BaseTest):
     self.assertIsNone(evaluator.evaluate(0))
 
   def test_exact__given_number_values__calls_is_finite_number(self):
-    """ Returns True if is_finite_number returns True. Returns None if is_finite_number returns False. """
+    """ Test that CustomAttributeConditionEvaluator.evaluate returns True
+        if is_finite_number returns True. Returns None if is_finite_number returns False. """
 
     evaluator = condition_helper.CustomAttributeConditionEvaluator(
         exact_int_condition_list, {'lasers_count': 9000}
       )
 
     with mock.patch('optimizely.helpers.validator.is_finite_number',
-                    return_value=True) as is_finite:
+                    return_value=True) as mock_is_finite:
       self.assertTrue(evaluator.evaluate(0))
-    is_finite.assert_called_with(9000)
+    mock_is_finite.assert_called_with(9000)
 
     evaluator = condition_helper.CustomAttributeConditionEvaluator(
       exact_int_condition_list, {'lasers_count': 9000.0}
     )
 
     with mock.patch('optimizely.helpers.validator.is_finite_number',
-                    return_value=False) as is_finite:
+                    return_value=False) as mock_is_finite:
       self.assertIsNone(evaluator.evaluate(0))
-    is_finite.assert_called_with(9000.0)
+    mock_is_finite.assert_called_with(9000.0)
 
   def test_exact_bool__returns_true__when_user_provided_value_is_equal_to_condition_value(self):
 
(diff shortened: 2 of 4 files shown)


oakbani:
Let me know if I should extend is_finite_check to revenue and numeric values within this PR. Looks like a PR on parsing revenue value is already pending for a while. https://github.com/optimizely/python-sdk/pull/134

optimizely/helpers/validator.py
```diff
@@ -185,7 +187,40 @@ def is_attribute_valid(attribute_key, attribute_value):
...
+  if not isinstance(value, (numbers.Integral, float)):
+      # numbers.Integral instead of int to accomodate long integer in python 2
+    return False
+
+  if isinstance(value, bool):
+    # bool is a subclass of int
+    return False
+
+  if isinstance(value, numbers.Integral):
+    if value > 1e53:
```
nchilada:
```suggestion
    if abs(value) > 1e53:
```

optimizely/helpers/validator.py
```diff
@@ -185,7 +187,40 @@ def is_attribute_valid(attribute_key, attribute_value):
...
+
+  if not isinstance(value, (numbers.Integral, float)):
+      # numbers.Integral instead of int to accomodate long integer in python 2
+    return False
+
+  if isinstance(value, bool):
+    # bool is a subclass of int
+    return False
+
+  if isinstance(value, numbers.Integral):
```
nchilada: We should apply the size limit even if the value is a float. Can we get rid of this gate?

optimizely/helpers/validator.py
```diff
@@ -185,7 +187,40 @@ def is_attribute_valid(attribute_key, attribute_value):
...
+  if isinstance(attribute_value, (string_types, bool)):
     return True
 
+  if isinstance(attribute_value, (numbers.Integral, float)):
+    return is_finite_number(attribute_value)
+
   return False
+
+
+def is_finite_number(value):
```
nchilada: Are you planning to merge this PR on its own and then update #146 to call `is_finite_number` during audience evaluation? That works for me!

tests/helpers_tests/test_validator.py
```diff
@@ -168,6 +171,55 @@ def test_is_attribute_valid(self):
...
+
+    with mock.patch('optimizely.helpers.validator.is_finite_number',
+                    return_value=False) as is_finite:
+      self.assertFalse(validator.is_attribute_valid('test_attribute', 5.5))
+
+    is_finite.assert_called_once_with(5.5)
+
+    with mock.patch('optimizely.helpers.validator.is_finite_number',
+                    return_value='abc') as is_finite:
+      self.assertEqual('abc', validator.is_attribute_valid('test_attribute', 0))
```
nchilada: Nit: I'm not sure this block is needed.

tests/helpers_tests/test_validator.py
```diff
@@ -168,6 +171,55 @@ def test_is_attribute_valid(self):
...
+    self.assertFalse(validator.is_finite_number([]))
+    self.assertFalse(validator.is_finite_number(()))
+
+    # test invalid numbers
+    self.assertFalse(validator.is_finite_number((float('inf'))))
+    self.assertFalse(validator.is_finite_number((float('-inf'))))
+    self.assertFalse(validator.is_finite_number((float('nan'))))
+    self.assertFalse(validator.is_finite_number(int(1e53) + 1))
+    if PY2:
+      self.assertFalse(validator.is_finite_number(long(1e53) + 1))
```
nchilada: As implied by my comments in the code under test, we also want assertions like these:
```python
self.assertFalse(validator.is_finite_number(float(1e53) + 1.0))
self.assertFalse(validator.is_finite_number(-int(1e53) - 1))
if PY2:
  self.assertFalse(validator.is_finite_number(-long(1e53) - 1))
self.assertFalse(validator.is_finite_number(-float(1e53) - 1.0))
```
(maybe reordered as you deem fit)

tests/helpers_tests/test_validator.py
```diff
@@ -168,6 +171,55 @@ def test_is_attribute_valid(self):
...
+
+    # test valid numbers
+    self.assertTrue(validator.is_finite_number((0)))
+    self.assertTrue(validator.is_finite_number((5)))
+    self.assertTrue(validator.is_finite_number((5.5)))
+    self.assertTrue(validator.is_finite_number((float(1e53) + 1)))
+    self.assertTrue(validator.is_finite_number((int(1e53))))
+    if PY2:
+      self.assertTrue(validator.is_finite_number((long(1e53))))
```
nchilada: And
```python
self.assertTrue(validator.is_finite_number((float(1e53))))
```

tests/helpers_tests/test_validator.py
```diff
@@ -168,6 +171,55 @@ def test_is_attribute_valid(self):
...
+    self.assertFalse(validator.is_finite_number('HelloWorld'))
+    self.assertFalse(validator.is_finite_number(True))
+    self.assertFalse(validator.is_finite_number(False))
+    self.assertFalse(validator.is_finite_number(None))
+    self.assertFalse(validator.is_finite_number({}))
+    self.assertFalse(validator.is_finite_number([]))
+    self.assertFalse(validator.is_finite_number(()))
+
+    # test invalid numbers
+    self.assertFalse(validator.is_finite_number((float('inf'))))
```
nchilada: Nit: there are unnecessary double parens in a lot of these assertions!
```suggestion
    self.assertFalse(validator.is_finite_number(float('inf')))
```

optimizely/helpers/validator.py
nchilada: Ugh - I just realized that I asked for the limit to be `1e53`/`10^53`, but it should actually be `2^53`. Do you think you could make that change too?

optimizely/helpers/validator.py
oakbani: Yeah. Once this is approved. I will merge it changes to #146

nchilada: Oops, looks like #146 is merged! But it actually added the calls to `is_finite_number`. So I guess you just need to rebase this branch onto devel in order to add the 2^53/NaN/Infinity limits.

nchilada:
> > Let me know if I should extend is_finite_check to revenue and numeric values within this PR. Looks like a PR on parsing revenue value is already pending for a while. #134
> 
> We can only effectively accept integers for `revenue`, so I'm thinking we should probably just close #134. 😕
> 
> I'm asking offline to figure out the exact size limits for an event's `revenue`/`value`/tags. Sorry for not having an immediate answer!

Let's skip this after all; it's taking a while to get an answer, and it's not directly related to our current epic, and it looks like we're skipping this in other SDKs (e.g. optimizely/objective-c-sdk#348).

tests/helpers_tests/test_validator.py
```diff
@@ -168,6 +171,61 @@ def test_is_attribute_valid(self):
...
+    self.assertFalse(validator.is_finite_number(-float(2**53) - 2.0))
+    if PY2:
+      self.assertFalse(validator.is_finite_number(long(2**53) + 1))
+      self.assertFalse(validator.is_finite_number(-long(2**53) - 1))
+
+    # test valid numbers
+    self.assertTrue(validator.is_finite_number(0))
+    self.assertTrue(validator.is_finite_number(5))
+    self.assertTrue(validator.is_finite_number(5.5))
+    self.assertTrue(validator.is_finite_number(float(2**53) + 1.0))
```
oakbani: https://stackoverflow.com/questions/1848700/biggest-integer-that-can-be-stored-in-a-double

tests/helpers_tests/test_condition.py
```diff
@@ -371,6 +429,42 @@ def test_greater_than_int__returns_true__when_user_value_greater_than_condition_
...
+    if PY2:
+      evaluator = condition_helper.CustomAttributeConditionEvaluator(
+        gt_int_condition_list, {'meters_travelled': long(49)}
+      )
+
+      with mock.patch('optimizely.helpers.validator.is_finite_number',
+                       side_effect=side_effect) as is_finite:
+        self.assertIsNone(evaluator.evaluate(0))
+      is_finite.assert_called_with(long(49))
+
```
nchilada: Ooh, I like the idea of using `side_effect` here. Can we go farther with this and test that `is_finite` only needs to reject the condition value _or_ the user attribute value in order to stop condition evaluation?

Something like
```python
def test_greater_than__calls_is_finite_number(self):
  evaluator = condition_helper.CustomAttributeConditionEvaluator(
    gt_int_condition_list, {'meters_travelled': 48.1}
  )
  
  def is_finite_number__rejecting_condition_value(value):
    if value == 48:
      return False
    return True
  
  with mock.patch('optimizely.helpers.validator.is_finite_number',
                  side_effect=is_finite_number__rejecting_condition_value):
    self.assertIsNone(evaluator.evaluate(0))
  
  def is_finite_number__rejecting_user_attribute_value(value):
    if value == 48.1:
      return False
    return True
  
  with mock.patch('optimizely.helpers.validator.is_finite_number',
                  side_effect=is_finite_number__rejecting_user_attribute_value):
    self.assertIsNone(evaluator.evaluate(0))
  
  def is_finite_number__accepting_both_values(value):
    return True
  
  with mock.patch('optimizely.helpers.validator.is_finite_number',
                  side_effect=is_finite_number__accepting_both_values):
    self.assertIsNone(evaluator.evaluate(0))
```

I don't think we need to test ints, longs, and floats separately... not here, at least... we can probably get away with `test_greater_than__calls_is_finite_number`, `test_less_than__calls_is_finite_number`, and maybe ``test_exact__given_number_values__calls_is_finite_number`. If that makes sense.

@oakbani @aliabbasrizvi what do you think?

oakbani: Nice suggestion. Revised accordingly.

optimizely/helpers/validator.py
```diff
@@ -187,20 +187,24 @@ def is_attribute_valid(attribute_key, attribute_value):
...
-  """ Method to validate if the given value is a number and not one of NAN, INF, -INF.
+  """ Validates if the given value is a number, enforces
+   absolute limit of 2^53 and restricts NAN, INF, -INF.
 
   Args:
     value: Value to be validated.
 
   Returns:
-    Boolean: True if value is a number and not NAN, INF or -INF else False.
+    Boolean: True if value is a number and not NAN, INF, -INF or greater than 2^53 else False.
```
msohailhussain: nit: or absolute limit of 2^53

aliabbasrizvi: Late to the party, but wouldn't I want this limit for Double type variables and Integer (sys.maxint) for Integer variables?

aliabbasrizvi: @nchilada ^

aliabbasrizvi: I think I get it. I am mixing feature variables here. Apologies.

nchilada: @aliabbasrizvi it's true that Python integers can have larger values than 2^53, but they're passed to the backend as JSON Numbers which are best limited to 2^53. Does that clarify why we limit integers too?

tests/helpers_tests/test_condition.py
```diff
@@ -290,19 +290,27 @@ def test_exact__given_number_values__calls_is_finite_number(self):
         exact_int_condition_list, {'lasers_count': 9000}
       )
 
+    # assert that isFiniteNumber only needs to reject condition value to stop evaluation.
     with mock.patch('optimizely.helpers.validator.is_finite_number',
-                    return_value=True) as mock_is_finite:
-      self.assertTrue(evaluator.evaluate(0))
-    mock_is_finite.assert_called_with(9000)
+                    side_effect=[False, True]) as mock_is_finite:
```
oakbani: @nchilada In exact match need to send exactly same values to ensure that it's the mocked is_finite that is really driving the result. Therefore, can't make a value-base side-effect method as in gt/lt methods. 
This returns False on 1st call and True on 2nd.

nchilada: Ah, you're right. Good solution. Thanks @oakbani!

PR #162, opened 2019-01-11
ci: Replace pep8 by flake8 and separate linting stage on Travis

oakbani:
Summary
-------
This implements flake8 which is a wrapper over pyflakes, pycodestyle(PEP8), and circular complexity checks. 

Also, the linting stage has been separated so that it immediately fails the build rather than creating separate environments for multiple python versions and failing all the jobs for the same reason. 

Test plan
---------

Issues
------

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,14 +8,13 @@ python:
   - "pypy"
   - "pypy3"
 install: "pip install -r requirements/core.txt;pip install -r requirements/test.txt"
-before_script: skip
 addons:
   srcclr: true
 script: "nosetests --with-coverage --cover-package=optimizely"
 after_success:
   - coveralls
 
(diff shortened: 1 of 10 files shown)


.travis.yml
```diff
@@ -8,7 +8,7 @@ python:
   - "pypy"
   - "pypy3"
 install: "pip install -r requirements/core.txt;pip install -r requirements/test.txt"
-before_script: "pep8"
+before_script: skip
```
aliabbasrizvi: nit. This can be removed right?

.travis.yml
```diff
@@ -17,12 +17,21 @@ after_success:
 
 # Integration tests need to run first to reset the PR build status to pending
```
msohailhussain: Please change this comment.

PR #176, opened 2019-05-15
Refactoring and passing ProjectConfig from Optimizely itself.

aliabbasrizvi:
Summary
-------

- Refactoring to pass around `ProjectConfig` vs every class having its own copy of `ProjectConfig`. This way `Bucketer`, `DecisionService` and `EventBuilder` are static and makes way for us to introduce datafile manager.

Test plan
---------

Unit tests and compatibility suite tests continue to pass.

### DIFF
--- a/tests/test_optimizely.py
+++ b/tests/test_optimizely.py
@@ -251,7 +251,7 @@ class OptimizelyTest(base.BaseTest):
       'revision': '42'
     }
     mock_decision.assert_called_once_with(
-      self.project_config.get_experiment_from_key('test_experiment'), 'test_user', None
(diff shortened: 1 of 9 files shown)


tests/test_optimizely.py
```diff
@@ -1960,23 +1963,23 @@ def test_get_enabled_features__broadcasts_decision_for_each_feature(self):
     mock_variation_2 = opt_obj.config.get_variation_from_id('test_experiment', '111128')
 
     def side_effect(*args, **kwargs):
-      feature = args[0]
+      feature = args[1]
```
aliabbasrizvi: This had to change as `project_config` is first param (index 0) now.

PR #177, opened 2019-05-16
feat(datafile-management): Introducing config manager for managing project config

aliabbasrizvi:
### Summary
This PR introduces config manager which polls for datafile after some interval and prepares project config based on it if it has updated.

### Tests
- Manual testing
- Unit tests


optimizely/config_manager.py
```diff
@@ -0,0 +1,198 @@
...
+    response = requests.get(self.datafile_url, headers=request_headers)
+    self.set_config(response)
+
+  def _run(self):
+    """ Triggered as part of the thread which fetches the datafile and sleeps until next update interval. """
+    while self.is_running:
+      self.fetch_datafile()
+      time.sleep(self.update_interval)
+
+  def start(self):
```
mikeng13: I think both `start` and `stop` should be methods on the base manager because I can see it being implemented in other types of implementations like push-based ones. In the static one, they can just be No-Op. I can see the Optimizely SDK, which will use this manager, would want to call `start`/`close` on this irrespective of the impl used.

optimizely/config_manager.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+import http
+import requests
+import threading
+import time
+from requests import exceptions as requests_exceptions
+
+from optimizely import exceptions
+from optimizely import logger as _logging
```
mikeng13: nit: can this be `optimizely_logger` instead of `logging`?

optimizely/config_manager.py
```diff
@@ -0,0 +1,198 @@
...
+                    SDK key to determine URL from which to fetch the datafile.
+
+    Returns:
+      String representing URL to fetch datafile from.
+    """
+    # Ensure that either is provided by the user.
+    if sdk_key is None and url is None:
+      raise exceptions.InvalidInputException('Must provide at least one of sdk_key or url.')
+
+    url_template = url_template or enums.ConfigManager.DATAFILE_URL_TEMPLATE
```
mikeng13: Does it make sense to just add `enums.ConfigManager.DATAFILE_URL_TEMPLATE` as a default value for the arg being passed in?

optimizely/config_manager.py
```diff
@@ -0,0 +1,198 @@
...
+
+    return url
+
+  def get_update_interval(self, update_interval):
+    """ Helper method to determine frequency at which datafile has to be polled and ProjectConfig updated.
+
+    Args:
+      update_interval: Time in seconds optionally sent in by the user.
+    """
+    polling_interval = update_interval or enums.ConfigManager.DEFAULT_UPDATE_INTERVAL
```
mikeng13: Why introduce a different name here? Why not continue calling it `update_interval`?

optimizely/config_manager.py
```diff
@@ -0,0 +1,198 @@
...
+
+  def fetch_datafile(self):
+    """ Fetch datafile and set ProjectConfig. """
+
+    request_headers = {}
+    if self.last_modified:
+      request_headers[enums.HTTPHeaders.IF_MODIFIED_SINCE] = self.last_modified
+
+    response = requests.get(self.datafile_url, headers=request_headers)
+    self.set_config(response)
```
mikeng13: `handle_response`? Seems like it does more than just set the config

optimizely/config_manager.py
```diff
@@ -0,0 +1,198 @@
+# Copyright 2019, Optimizely
```
mikeng13: Can we split this into different files? Ideally `BaseConfigManager` is it's own so that users looking to write their own impl can just import that and not have to import all these other implementations that we offer

optimizely/config_manager.py
```diff
@@ -0,0 +1,198 @@
...
+    if self.last_modified:
+      request_headers[enums.HTTPHeaders.IF_MODIFIED_SINCE] = self.last_modified
+
+    response = requests.get(self.datafile_url, headers=request_headers)
+    self.set_config(response)
+
+  def _run(self):
+    """ Triggered as part of the thread which fetches the datafile and sleeps until next update interval. """
+    while self.is_running:
+      self.fetch_datafile()
```
mjc1283: Nit: if we want this to be consistent with other SDKs, it should wait for the previous request to complete before starting the next one.

optimizely/config_manager.py
aliabbasrizvi: Yeah, will change that.

optimizely/config_manager.py
aliabbasrizvi: Yeah, ideally. I had been going back and forth on how to organize this. Will think more about this.

@travisbeck @nickdirienzo @onufryk any thoughts?

optimizely/config_manager.py
aliabbasrizvi: Sure. Will update.

optimizely/config_manager.py
aliabbasrizvi: True, will update.

optimizely/config_manager.py
aliabbasrizvi: Here `fetch_datafile` will complete before hitting going to the next step.

optimizely/config_manager.py
aliabbasrizvi: I was going by the interface we arrived on when implementing the Java SDK. I think at its core, the config manager should be responsible for returning the config through a `get_config` method. 
Building start/stop should not be necessary to be able to use it. In the current work, we can actually start the thread in constructor and not worry about stop as when it goes out of scope, the thread will be stopped anyway.

optimizely/config_manager.py
mjc1283: Ah, right. Then I think this implementation has scheduling behavior like JS used to have in this slide:

https://docs.google.com/presentation/d/1d6B3VxZLU06wsg3RPBqZ1vJImvf3BHCnbbtsJUbdKgY/edit#slide=id.p

The timer is not started until the response, but we wanted to standardize on the Java behavior.

mjc1283: It could be that making this change is not worth the extra complexity. I am still nit picking.

optimizely/config_manager.py
aliabbasrizvi: Addressed it in caller itself.

optimizely/config_manager.py
aliabbasrizvi: Synced offline about this. For this implementation we are planning to keep it like this.

optimizely/config_manager.py
```diff
@@ -0,0 +1,198 @@
...
+      logger: Provides a logger instance.
+      error_handler: Provides a handle_error method to handle exceptions.
+    """
+    self.datafile_url = self.get_datafile_url(sdk_key, url, url_template)
+    self.update_interval = self.get_update_interval(update_interval)
+    self.last_modified = None
+    self._datafile = None
+    self._config = None
+    self._polling_thread = threading.Thread(target=self._run)
+    self.is_running = False
```
nickdirienzo: Instead of managing our own `is_running`, should we delegate that to `self._polling_thread.isAlive()`? 

https://docs.python.org/3/library/threading.html#threading.Thread.is_alive

This class could expose an `@property` for that:

```
@property
def is_running(self):
  return self._polling_thread.isAlive()
```

I believe the semantics are the same, but worth double-checking my understanding.

optimizely/config_manager.py
nickdirienzo: I'm personally okay if they're in the same module, as the implementations seem pretty standard and  the module is fairly small (< 200 lines). I'm also okay if you want to split the base class out, but I don't find it necessary. Just my 2c. :)

optimizely/config_manager.py
```diff
@@ -0,0 +1,198 @@
...
+
+
+class StaticConfigManager(BaseConfigManager):
+  """ Config manager that returns ProjectConfig based on provided datafile. """
+
+  def __init__(self,
+               datafile,
+               logger=None,
+               error_handler=None):
+    self.logger = logger or _logging.adapt_logger(logger or _logging.NoOpLogger())
```
nickdirienzo: Can we add a docstring for what these arguments are?

optimizely/config_manager.py
aliabbasrizvi: Nice. Will take a look.

optimizely/config_manager.py
aliabbasrizvi: Will add.

optimizely/config_manager.py
```diff
@@ -0,0 +1,198 @@
...
+from requests import exceptions as requests_exceptions
+
+from optimizely import exceptions
+from optimizely import logger as _logging
+from optimizely import project_config
+from optimizely.error_handler import NoOpErrorHandler as noop_error_handler
+from optimizely.helpers import enums
+
+
+class BaseConfigManager(object):
```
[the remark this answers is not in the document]
aliabbasrizvi: Yes, I did and I think I used it previously. Going to use that. Thanks.

optimizely/config_manager.py
```diff
@@ -0,0 +1,198 @@
...
+    """ Get config for use by optimizely.Optimizely.
+    The config should be an instance of project_config.ProjectConfig."""
+    raise NotImplementedError
+
+
+class StaticConfigManager(BaseConfigManager):
+  """ Config manager that returns ProjectConfig based on provided datafile. """
+
+  def __init__(self,
+               datafile,
```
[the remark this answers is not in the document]
aliabbasrizvi: Oh. Good question. The idea is that when you initialize Optimizely:
1. If only SDK key is provided use `PollingConfigManager`.
2. If only datafile is provided use `StaticConfigManager`.
3. If both are provided use `PollingConfigManager` initialized with datafile (I realize I need to add it to the init method).

Let me know what you think.

optimizely/config_manager.py
```diff
@@ -0,0 +1,198 @@
...
+    Returns:
+      ProjectConfig.
+    """
+    return self._config
+
+
+class PollingConfigManager(BaseConfigManager):
+  """ Config manager that polls for the datafile and updated ProjectConfig based on an update interval. """
+
+  def __init__(self,
```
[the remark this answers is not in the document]
aliabbasrizvi: Good call. Need to add `datafile` here as well. See comment above in response to your earlier question.

optimizely/config_manager.py
```diff
@@ -0,0 +1,198 @@
...
+  def _run(self):
+    """ Triggered as part of the thread which fetches the datafile and sleeps until next update interval. """
+    while self.is_running:
+      self.fetch_datafile()
+      time.sleep(self.update_interval)
+
+  def start(self):
+    """ Start the config manager and the thread to periodically fetch datafile. """
+    if not self.is_running:
+      self.is_running= True
```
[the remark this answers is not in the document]
aliabbasrizvi: oops.

optimizely/config_manager.py
```diff
@@ -0,0 +1,233 @@
...
+        self.set_config(response.text)
+
+    def fetch_datafile(self):
+        """ Fetch datafile and set ProjectConfig. """
+
+        request_headers = {}
+        if self.last_modified:
+            request_headers[enums.HTTPHeaders.IF_MODIFIED_SINCE] = self.last_modified
+
+        response = requests.get(self.datafile_url, headers=request_headers)
```
oakbani: @aliabbasrizvi  Should we be passing timeout here? If no timeout is specified, it may hang this thread. https://2.python-requests.org//en/master/user/quickstart/#timeouts

aliabbasrizvi: Yes. You are right. Will update.

optimizely/config_manager.py
```diff
@@ -0,0 +1,233 @@
...
+ABC = abc.ABCMeta('ABC', (object,), {'__slots__': ()})
+
+
+class BaseConfigManager(ABC):
+    """ Base class for Optimizely's config manager. """
+
+    def __init__(self,
+                 logger=None,
+                 error_handler=None,
+                 **kwargs):
```
nickdirienzo: Should we accept arbitrary `kwargs` on the base class?

optimizely/config_manager.py
```diff
@@ -0,0 +1,233 @@
...
+        if url is None:
+            try:
+                return url_template.format(sdk_key=sdk_key)
+            except (AttributeError, KeyError):
+                raise optimizely_exceptions.InvalidInputException(
+                    'Invalid url_template {} provided.'.format(url_template))
+
+        return url
+
+    def set_update_interval(self, update_interval):
```
nickdirienzo: I believe we're missing a docstring for `update_interval` here?

optimizely/config_manager.py
```diff
@@ -0,0 +1,233 @@
...
+        """ Returns instance of ProjectConfig.
+
+        Returns:
+          ProjectConfig.
+        """
+        return self._config
+
+
+class PollingConfigManager(BaseConfigManager):
+    """ Config manager that polls for the datafile and updated ProjectConfig based on an update interval. """
```
nickdirienzo: This applies both here and for `StaticConfigManager`: should we have a docstring example of how to use this class? The tests do a good job, but I'm wondering where we will point developers for an example of how to use it in practice.

optimizely/config_manager.py
```diff
@@ -0,0 +1,233 @@
...
+                 update_interval=None,
+                 url=None,
+                 url_template=None,
+                 logger=None,
+                 error_handler=None):
+        """ Initialize config manager. One of sdk_key or url has to be set to be able to use.
+
+        Args:
+          sdk_key: Optional string uniquely identifying the datafile.
+          update_interval: Optional floating point number representing time interval in seconds
```
mikeng13: nit: docstring missing for `datafile`

optimizely/config_manager.py
```diff
@@ -0,0 +1,233 @@
...
+        """ Check if polling thread is alive or not. """
+        return self._polling_thread.is_alive()
+
+    def _run(self):
+        """ Triggered as part of the thread which fetches the datafile and sleeps until next update interval. """
+        while self.is_running:
+            self.fetch_datafile()
+            time.sleep(self.update_interval)
+
+    def start(self):
```
mikeng13: We also need a way to stop the updates

optimizely/config_manager.py
```diff
@@ -0,0 +1,233 @@
...
+        """ Looks up and sets datafile and config based on response body.
+
+         Args:
+           datafile: JSON string representing the Optimizely project.
+         """
+        # TODO(ali): Add validation here to make sure that we do not update datafile and config if not a valid datafile.
+        self._datafile = datafile
+        # TODO(ali): Add notification listener.
+        self._config = project_config.ProjectConfig(self._datafile, self.logger, self.error_handler)
+        self.logger.info('Received new datafile and updated config.')
```
mikeng13: I think this could just be a `debug`. Info is kinda noisy if we expect datafile updates to be happening recurringly

optimizely/config_manager.py
aliabbasrizvi: Not really. Removed.

optimizely/config_manager.py
aliabbasrizvi: Yes. Good catch.

optimizely/config_manager.py
aliabbasrizvi: Added

optimizely/config_manager.py
aliabbasrizvi: Good question. I think we would want to have good documentation on developers.optimizely.com and/or add a usage section in README.md explaining everything about which class to use and when.

optimizely/config_manager.py
aliabbasrizvi: Because we making `polling_thread` a daemon thread, it stops when the main thread stops: https://docs.python.org/2/library/threading.html#threading.Thread.daemon and so we do not need a stop. It auto-stops. Also, I will be calling this start in the constructor itself when I integrate this with optimizely.Optimizely.

optimizely/config_manager.py
nickdirienzo: I like both of these ideas! Should usage be included in this PR?

nickdirienzo:
Although, we shouldn't merge as it looks like this change is breaking 3.4+. It looks like there's a TypeError. Python 3 is more strict with string types than Python 2:

```
Traceback (most recent call last):
  File "/home/travis/build/optimizely/python-sdk/tests/test_config_manager.py", line 149, in test_fetch_datafile
    project_config_manager.fetch_datafile()
  File "/home/travis/build/optimizely/python-sdk/optimizely/config_manager.py", line 223, in fetch_datafile
    self._handle_response(response)
  File "/home/travis/build/optimizely/python-sdk/optimizely/config_manager.py", line 211, in _handle_response
    self.set_config(response.text)
  File "/home/travis/virtualenv/python3.4.6/lib/python3.4/site-packages/requests/models.py", line 857, in text
    encoding = self.apparent_encoding
  File "/home/travis/virtualenv/python3.4.6/lib/python3.4/site-packages/requests/models.py", line 727, in apparent_encoding
    return chardet.detect(self.content)['encoding']
  File "/home/travis/virtualenv/python3.4.6/lib/python3.4/site-packages/chardet/__init__.py", line 34, in detect
    '{0}'.format(type(byte_str)))
TypeError: Expected object of type bytes or bytearray, got: <class 'str'>
```

I believe you should be able to do `str.encode(response.text)` in `config_manager.py`. That should work in both Python 2 and Python 3.

In Python 2:
```
Python 2.7.15 (default, Aug 22 2018, 16:36:18)
[GCC 4.2.1 Compatible Apple LLVM 9.1.0 (clang-902.0.39.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> str.encode
<method 'encode' of 'str' objects>
>>> type(str.encode('asf'))
<type 'str'>
```

In Python 3:
```
Loading .env environment variables…
Python 3.7.0 (default, Jun 29 2018, 20:14:27)
[Clang 9.0.0 (clang-900.0.39.2)] on darwin
Type "help", "copyright", "credits" or "license" for more information.
>>> type(str.encode('asf'))
<class 'bytes'>
```

optimizely/config_manager.py
aliabbasrizvi: Will add as a follow up.

nickdirienzo:
Oh I was quite wrong. Sorry about that, @aliabbasrizvi. It looks like [`Response.content`](https://2.python-requests.org/en/master/api/#requests.Response.content) will always return `bytes`. Perhaps this is what we want to make it 2/3 compatible?

PR #179, opened 2019-06-05
feat(datafile-management): Integrate config_manager with Optimizely

aliabbasrizvi:
Summary
-------
In this change, we move to using `config_manager.PollingConfigManager` and `config_manager.StaticConfigManager` based on how the SDK is initialized. With the change the SDK may be initialized by either:

- Providing `sdk_key` only: in this case, the `PollingConfigManager` will be used which will request and update config at default interval of 5 minutes.
- Providing `datafile` only: in this case, the `StaticConfigManager` will be used and the SDK will continue to behave like it has done previously i.e. it will need to be re-initialized.
- Providing both options: in this case, the `PollingConfigManager` will be used and the SDK will start with a config based on the provided datafile and will update config at default interval of 5 minutes.

To use more options on the `PollingConfigManager` like different `update_interval`, `url`, `url_template`, users will have the option to create an instance of `PollingConfigManager` and pass it to `Optimizely` as `config_manager` option.

Test plan
---------
- Manual testing
- Unit tests


aliabbasrizvi:
Not ready for review yet.

optimizely/optimizely.py
```diff
@@ -63,51 +70,41 @@ def __init__(self,
...
-    finally:
-      if error_msg:
-        self.is_valid = False
-        # We actually want to log this error to stderr, so make sure the logger
-        # has a handler capable of doing that.
-        self.logger = _logging.reset_logger(self.logger_name)
-        self.logger.exception(error_msg)
-        self.error_handler.handle_error(error_to_handle)
-        return
+    if not self.config_manager:
```
mikeng13: Does it make sense to combine this with the `sdk_key` check? There doesn't seem to be additional logic outside of the inner `if` clause

tests/helpers_tests/test_validator.py
```diff
@@ -26,6 +27,21 @@
 
 class ValidatorTest(base.BaseTest):
 
+  def test_is_config_manager_valid__returns_true(self):
+    """ Test that valid config_manager returns True. """
```
mikeng13: nit: `... returns True for valid config manager implementations.`

tests/helpers_tests/test_validator.py
```diff
@@ -26,6 +27,21 @@
...
+
+    self.assertTrue(validator.is_config_manager_valid(config_manager.StaticConfigManager))
+    self.assertTrue(validator.is_config_manager_valid(config_manager.PollingConfigManager))
+
+  def test_is_config_manager_valid__returns_false(self):
+    """ Test that invalid config_manager returns False. """
```
mikeng13: ditto above

optimizely/helpers/enums.py
```diff
@@ -71,18 +71,18 @@ class DecisionSources(object):
 
 
 class Errors(object):
-  INVALID_ATTRIBUTE_ERROR = 'Provided attribute is not in datafile.'
```
mikeng13: much more concise :)

optimizely/config_manager.py
```diff
@@ -56,29 +56,67 @@ class StaticConfigManager(BaseConfigManager):
...
+            error_msg = enums.Errors.INVALID_INPUT.format('datafile')
+            error_to_handle = optimizely_exceptions.InvalidInputException(error_msg)
+        finally:
+            if error_msg:
+                self.logger.error(error_msg)
+                self.error_handler.handle_error(error_to_handle)
+                return
+
+        # TODO(ali): Add notification listener.
+        self._config = config
```
mikeng13: We need to check whether the revision is different before setting it.

aliabbasrizvi: We don't do so in other SDKs and I think we should respect the last modified timestamp (which is happening already).

optimizely/helpers/enums.py
aliabbasrizvi: Yeah, the last part was redundant.

optimizely/config_manager.py
mikeng13: Ah you're right. The timestamp already takes care of that.

optimizely/optimizely.py
```diff
@@ -63,51 +70,41 @@ def __init__(self,
...
+      if sdk_key:
+        self.config_manager = PollingConfigManager(sdk_key=sdk_key,
```
mjc1283: If the user wants to provide custom `url_template` or `update_interval`, do they need to construct a config manager outside using those options, then pass it in to `Optimizely`?

optimizely/config_manager.py
mjc1283: The config could be accessed by different threads, so do we need to do anything to account for this? For example, in Java, the [currentProjectConfig variable is an AtomicReference](https://github.com/optimizely/java-sdk/blob/c7422462f5a56b3bfb796abcd543bfcc3bac3d57/core-api/src/main/java/com/optimizely/ab/config/PollingProjectConfigManager.java#L45).

aliabbasrizvi: Isn't check on entire content a better check? Also, you are not checking if new version is higher. I feel in JS SDK you have already gone through the entire effort and then throwing away the config. There is a new datafile available and the config manager should just respond to that by updating.
What do you think?

optimizely/config_manager.py
```diff
@@ -56,29 +56,67 @@ class StaticConfigManager(BaseConfigManager):
...
+        self.logger.debug('Received new datafile and updated config.')
```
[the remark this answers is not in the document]
aliabbasrizvi: Yes. We can do that.

optimizely/config_manager.py
```diff
@@ -56,29 +56,67 @@ class StaticConfigManager(BaseConfigManager):
...
 
         Returns:
-          ProjectConfig.
+            ProjectConfig. None if not set.
         """
         return self._config
 
 
-class PollingConfigManager(BaseConfigManager):
+class PollingConfigManager(StaticConfigManager):
```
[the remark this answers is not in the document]
aliabbasrizvi: I think you are right. `set_config` should be marked private and not be used.
`PollingConfigManager` sublassing `StaticConfigManager` made sense as they do exactly the same thing. 

I don't think config manager interface (captured in `BaseConfigManager`) should have `set_config`. It should only support `get_config`.

optimizely/optimizely.py
aliabbasrizvi: Yes. That is the expectation currently. I was worried about proliferation of options on `optimizely.Optimizely` and so I feel if users really want to use those options they can just create `config_manager` and pass it in.

What do you think?

optimizely/config_manager.py
aliabbasrizvi: There is only 1 thread at play here (_polling_thread). Nothing else will access this.

optimizely/optimizely.py
mjc1283: Sounds good to me :+1:

optimizely/config_manager.py
aliabbasrizvi: Updating this.

PR #182, opened 2019-06-18
Fixing multiple issues in notification_center.

aliabbasrizvi:
Summary
-------

`notification_center.NotificationCenter` did not have any unit tests for the class. All the related unit tests were absorbed in tests for `optimizely.Optimizely`. This PR:
- Creates `test_notification_center.py` to introduce unit tests for the `NotificationCenter` class.
- Fixes naming in the `NotificationCenter` class. This includes fixing incorrect API names which have been marked as deprecated and we will discontinue support in the next major release.
- Removes tests in `test_optimizely.py` which were focusing on tests captured in `test_notification_center.py`.

Test plan
---------

- Unit tests

### DIFF
--- a/optimizely/notification_center.py
+++ b/optimizely/notification_center.py
@@ -1,4 +1,4 @@
-# Copyright 2017, Optimizely
+# Copyright 2017-2019, Optimizely
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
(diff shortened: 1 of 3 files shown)


optimizely/notification_center.py
```diff
@@ -1,4 +1,4 @@
-# Copyright 2017, Optimizely
+# Copyright 2017-2019, Optimizely
```
mikeng13: Has it really been 2 years since we touched this?

optimizely/notification_center.py
```diff
@@ -14,43 +14,52 @@
...
-                filter(lambda tup: tup[1] == notification_callback, self.notifications[notification_type]),
-                0) > 0:
-          return -1
-      self.notifications[notification_type].append((self.notification_id, notification_callback))
+    if notification_type not in NOTIFICATION_TYPES:
+      self.logger.error('Invalid notification_type: {} provided. Not adding listener.'.format(notification_type))
+      return -1
 
-    ret_val = self.notification_id
+    if reduce(lambda a, b: a + 1, filter(lambda tup: tup[1] == notification_callback,
```
mikeng13: I don't find this to be very readable, but then again, I am not fluent in Python :)

optimizely/notification_center.py
aliabbasrizvi: We touched it some time last year briefly, but this was not updated.

optimizely/notification_center.py
aliabbasrizvi: Let me try and update this as well. This entire thing was not very readable.

PR #185, opened 2019-07-02
Handling update_interval values.

aliabbasrizvi:
Summary
-------
Handling `update_interval` values if they cause overflow or are of the invalid type.

### DIFF
--- a/optimizely/config_manager.py
+++ b/optimizely/config_manager.py
@@ -212,15 +212,24 @@ class PollingConfigManager(StaticConfigManager):
         Args:
           update_interval: Time in seconds after which to update datafile.
         """
-        self.update_interval = update_interval or enums.ConfigManager.DEFAULT_UPDATE_INTERVAL
+        if not update_interval:
+            update_interval = enums.ConfigManager.DEFAULT_UPDATE_INTERVAL
+            self.logger.debug('Set config update interval to default value {}.'.format(update_interval))
+
+        if not isinstance(update_interval, (int, float)):
+            raise optimizely_exceptions.InvalidInputException(
+                'Invalid update_interval "{}" provided.'.format(update_interval)
+            )
 
         # If polling interval is less than minimum allowed interval then set it to default update interval.
-        if self.update_interval < enums.ConfigManager.MIN_UPDATE_INTERVAL:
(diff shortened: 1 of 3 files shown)


optimizely/config_manager.py
```diff
@@ -269,9 +278,14 @@ def is_running(self):
...
-            self.fetch_datafile()
-            time.sleep(self.update_interval)
+        try:
+          while self.is_running:
+              self.fetch_datafile()
+              time.sleep(self.update_interval)
+        except (OSError, OverflowError) as err:
+            self.logger.error('Error in time.sleep. '
+                              'Provided update_interval value may be too big. Error: {}'.format(str(err)))
+            raise
```
mikeng13: Should we still be raising in this case or just pass it to the error handler? Not sure if raising from another thread might be an issue

aliabbasrizvi: I think if we hit either of these errors we should just raise so that the user sees this and fixes it. We cannot leave it silent.

mikeng13: Won't this crash the application though? The job of the error handler in this case would be raise visibility (if an error handler is provided), otherwise it'll just show up in the logs (again, if a logger is provided)

aliabbasrizvi: If the value (which is provided the first time) is too high to cause `OverflowError` or `OSError` the SDK won't be initialized in the first place.

mikeng13: I see. Wouldn't that make this check redundant then?

aliabbasrizvi: Yeah, it may seem that way, but unfortunately `time.sleep` has different limits. See here: https://github.com/python/cpython/blob/2.7/Modules/timemodule.c#L1030-L1155

We are not imposing any limit at init after our discussion last week. I was in favor on imposing a 1 day limit, but may be it is ok for `time.sleep` to fail to let people know that they should change it to a reasonable value. That way they can adjust for their platform accordingly.

mikeng13: Got it. Thanks for the explanation.

aliabbasrizvi: 👍

PR #191, opened 2019-07-22
Implement get_feature_variable and create unit tests

brandonhudavid:
## Summary
- Implement `get_feature_variable` method to return value of variable with any type 
- Create new unit tests to ensure functionality of `get_feature_variable`

Python is a dynamically-typed language, like [JavaScript](https://github.com/optimizely/javascript-sdk/pull/298), so it suffices to have a single method return the value of a feature variable rather than have a separate method for each possible feature variable type.

## Test plan
- Enable gherkin integration tests for Python in FSC suite and python-testapp

## Issues
- Fulfills one of the [SDK convenience APIs](https://docs.google.com/document/d/1MCsMioa4TNImryaDdOJJsZwL9j5zFvFIXODArHF3ORs/edit#) (design doc [here](https://docs.google.com/document/d/1CZ0iappDujwgQzYR8oiU9LRVzLho9jGQt2zaB1rpB48/edit#heading=h.6jynaot9cbnq))

### DIFF
--- a/optimizely/optimizely.py
+++ b/optimizely/optimizely.py
@@ -203,7 +203,7 @@ class Optimizely(object):
     if not variable:
       return None
 
-    # Return None if type differs
+    # For non-typed method, use type of variable; else, return None if type differs
     if not variable.type:
       variableType = variable.type
     elif variable.type != variable_type:
@@ -516,7 +516,7 @@ class Optimizely(object):
     return enabled_features
 
   def get_feature_variable(self, feature_key, variable_key, user_id, attributes=None):
-    """ Returns value for a certain boolean variable attached to a feature flag.
+    """ Returns value for a variable attached to a feature flag.
 
     Args:
       feature_key: Key of the feature whose variable's value is being accessed.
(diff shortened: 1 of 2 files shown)


optimizely/optimizely.py
```diff
@@ -203,8 +203,10 @@ def _get_feature_variable_for_type(self, feature_key, variable_key, variable_typ
     if not variable:
       return None
 
-    # Return None if type differs
-    if variable.type != variable_type:
+    # For non-typed method, use type of variable; else, return None if type differs
+    if not variable_type:
+      variable_type = variable.type
```
aliabbasrizvi: You can replace the 2 lines with:
```
variable_type = variable_tupe or variable.type
```

tests/test_optimizely.py
```diff
@@ -2256,124 +2391,423 @@ def test_get_feature_variable_integer_for_feature_in_rollout(self):
...
         'variable_key': 'count',
-        'variable_value': 399,
+        'variable_value': 399,
+        'variable_type': 'integer',
+        'source_info': {}
+      }
+    )
+
+  def test_get_feature_variable_string_for_feature_in_rollout(self):
+    """ Test that get_feature_variable_double returns Double value as expected \
```
aliabbasrizvi: Comment talks about double but test is for feature variable string API.

brandonhudavid: Updated comment to match its method.

optimizely/optimizely.py
brandonhudavid: Replaced the 2 lines with your suggestion. Neat trick.

PR #203, opened 2019-08-16
feat(eventProcessor): Add EventProcessor and BatchEventProcessor

mnoman09:
Summary
-------

- Introduces an EventProcessor interface.
- Introduces a BatchEventProcessor

Buffering events within a queue before dispatching is an optimization that should prevent SDK implementations from exhausting resources while increasing throughput. This implementation relies on a BlockingCollection to buffer events received from one-to-many producers. A single consumer thread continuously polls from this queue to build a batch before emitting the batched LogEvent.

Test plan
---------

- Added unit tests.


optimizely/closeable.py
```diff
@@ -0,0 +1,25 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+import abc
+
+ABC = abc.ABCMeta('ABC', (object,), {'__slots__': ()})
+
+
+class Closeable(object):
```
aliabbasrizvi: Why do we need this class? I don't want us to impose object oriented programming.

optimizely/config_manager.py
```diff
@@ -30,282 +30,282 @@
 
 
 class BaseConfigManager(ABC):
```
aliabbasrizvi: The 4 space indent was intentional. We want to move to that.

Please restore.

tox.ini
```diff
@@ -4,6 +4,7 @@
 # E121 - continuation line indentation is not a multiple of four
 # E127 - continuation line over-indented for visual indent
 # E722 - do not use bare 'except'
-ignore = E111,E114,E121,E127, E722
+# W504 - line break after binary operator
```
aliabbasrizvi: Remove this and fix the indent problem. We are trying to clean up the remaining rules and become flake8 compliant.

optimizely/config_manager.py
```diff
@@ -30,282 +30,282 @@
...
+     Args:
+       datafile: JSON string representing the Optimizely project.
+     """
+
+    if self.validate_schema:
+        if not validator.is_datafile_valid(datafile):
+            self.logger.error(enums.Errors.INVALID_INPUT.format('datafile'))
             return
 
-        self._config = config
```
aliabbasrizvi: How are config manager changes related to event processor changes? Please remove these changes from this PR.

optimizely/event/event_processor.py
```diff
@@ -0,0 +1,225 @@
...
+class EventProcessor(ABC):
+  """ Class encapsulating event_processor functionality. Override with your own processor
+  providing process method. """
+
+  @abc.abstractmethod
+  def process(user_event):
+    pass
+
+
+class BatchEventProcessor(EventProcessor, Closeable):
```
aliabbasrizvi: So here is what I would recommend. 

You should define a timer class that does stuff based on a defined interval. I am hoping that the timer class will be useful in both config manager and here in the batch event processor.

oakbani: do you mean something similar to this https://github.com/optimizely/ruby-sdk/blob/master/lib/optimizely/config_manager/async_scheduler.rb

optimizely/event/event_processor.py
```diff
@@ -0,0 +1,237 @@
...
+
+  def _validate_intantiation_props(self, prop, prop_name):
+    if (prop_name == 'batch_size' and not isinstance(prop, int)) or prop is None or prop < 1 or \
+      not validator.is_finite_number(prop):
+      self.logger.info('Using default value for {}.'.format(prop_name))
+      return False
+
+    return True
+
+  def _get_time(self, _time=None):
```
aliabbasrizvi: What is the purpose of this time? What is the unit? Can we add docstring?

optimizely/event/event_processor.py
```diff
@@ -0,0 +1,237 @@
...
+    if _time is None:
+      return int(round(time.time()))
+
+    return int(round(_time))
+
+  def start(self):
```
aliabbasrizvi: What is the purpose of this method? Can we add docstring?

optimizely/event/event_processor.py
```diff
@@ -0,0 +1,237 @@
...
+
+    self.flushing_interval_deadline = self._get_time() + self._get_time(self.flush_interval.total_seconds())
+    self.executor = threading.Thread(target=self._run)
+    self.executor.setDaemon(True)
+    self.executor.start()
+
+    self._is_started = True
+
+  def _run(self):
+    """ Scheduler method that periodically flushes events queue. """
```
aliabbasrizvi: This is not entirely true right?

optimizely/event/event_processor.py
```diff
@@ -0,0 +1,237 @@
...
+    try:
+      while True:
+        if self._get_time() > self.flushing_interval_deadline:
+          self._flush_queue()
+
+        try:
+          item = self.event_queue.get(True, 0.05)
+
+        except queue.Empty:
+          self.logger.debug('Empty queue, sleeping for 50ms.')
```
aliabbasrizvi: nit. This can be a very loud message happening 20 times per second. I'd recommend not having this message.

PR #208, opened 2019-09-05
fix(config-manager): Updated polling interval condition.

rashidsp:
Summary
-------

-  Update condition for using default update_interval when polling files.

### DIFF
--- a/optimizely/config_manager.py
+++ b/optimizely/config_manager.py
@@ -246,8 +246,8 @@ class PollingConfigManager(StaticConfigManager):
                 'Invalid update_interval "{}" provided.'.format(update_interval)
             )
 
-        # If polling interval is less than minimum allowed interval then set it to default update interval.
-        if update_interval < enums.ConfigManager.MIN_UPDATE_INTERVAL:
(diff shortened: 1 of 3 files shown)


msohailhussain:
Please update readme.md for valid polling interval.

msohailhussain:
Please update readme.md for polling interval.

optimizely/config_manager.py
```diff
@@ -246,8 +246,8 @@ def set_update_interval(self, update_interval):
                 'Invalid update_interval "{}" provided.'.format(update_interval)
             )
 
-        # If polling interval is less than minimum allowed interval then set it to default update interval.
-        if update_interval < enums.ConfigManager.MIN_UPDATE_INTERVAL:
+        # If polling interval is less than or equal to 0 then set it to default update interval.
+        if update_interval <= 0:
             self.logger.debug('update_interval value {} too small. Defaulting to {}'.format(
```
msohailhussain: `update_interval value is not greater than zero, Defaulting to {} value`. @aliabbasrizvi WDYT?

aliabbasrizvi: I think current message is fine.

PR #210, opened 2019-09-17
tests: Fix intemittent multithreaded unit test failure

oakbani:
Summary
-------
A Polling Manager unit test intermittently fails on some Python versions. 
See here: https://travis-ci.org/optimizely/python-sdk/jobs/585691111

Trying multiple approaches of using a short polling interval, using a blocking queue with a mocked side_effect, and putting a time sleep before assertion reveals that this is some issue with mocked object on pypy and pypy3 that it does not know even if it has been called many times. 

We can get rid of this assertion since we are already asserting is_running flag to ensure that the thread is running

Test plan
---------

Should pass existing unit tests. 

Issues
------

-  “THING-1234” or “Fixes #123”

### DIFF
--- a/optimizely/config_manager.py
+++ b/optimizely/config_manager.py
@@ -299,7 +299,6 @@ class PollingConfigManager(StaticConfigManager):
         try:
           while self.is_running:
               self.fetch_datafile()
-              self.logger.debug("This thread is running")
               time.sleep(self.update_interval)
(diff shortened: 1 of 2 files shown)


oakbani:
Looks like the issue with pypy/pypy3 is something else. The background thread is running as I can see the log but the mocked object 'fetch_datafile' is not being told that it has been called. 
image

PR #211, opened 2019-09-23
feat: Add blocking timeout in polling manager

rashidsp:
Summary
-------
This introduces blocking timeout in Python Config Manager. Config Manager will block the main thread until the config is available or the timeout occurs. 

https://github.com/optimizely/python-testapp/pull/63 should be merged to pass FSC tests for this PR.

### DIFF
--- a/optimizely/config_manager.py
+++ b/optimizely/config_manager.py
@@ -96,8 +96,7 @@ class StaticConfigManager(BaseConfigManager):
                                                   notification_center=notification_center)
         self._config = None
         self.validate_schema = not skip_json_validation
-        self.blocking_timeout = None
-        self._event = threading.Event()
+        self._configReadyEvent = threading.Event()
         self._set_config(datafile)
 
     def _set_config(self, datafile):
@@ -134,8 +133,9 @@ class StaticConfigManager(BaseConfigManager):
 
         if previous_revision == config.get_revision():
             return
+
         self._config = config
-        self._event.set()
+        self._configReadyEvent.set()
         self.notification_center.send_notifications(enums.NotificationTypes.OPTIMIZELY_CONFIG_UPDATE)
         self.logger.debug(
             'Received new datafile and updated config. '
@@ -148,8 +148,7 @@ class StaticConfigManager(BaseConfigManager):
         Returns:
             ProjectConfig. None if not set.
         """
-        if self.blocking_timeout is not None:
-            self._event.wait(self.blocking_timeout)
+
         return self._config
 
 
@@ -174,7 +173,7 @@ class PollingConfigManager(StaticConfigManager):
             datafile: Optional JSON string representing the project.
             update_interval: Optional floating point number representing time interval in seconds
                              at which to request datafile and set ProjectConfig.
-            blocking_timeout: Optional Time in seconds to block the config call until config object
+            blocking_timeout: Optional Time in seconds to block the get_config call until config object
                               has been initialized.
             url: Optional string representing URL from where to fetch the datafile. If set it supersedes the sdk_key.
             url_template: Optional string template which in conjunction with sdk_key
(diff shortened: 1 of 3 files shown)


optimizely/config_manager.py
```diff
@@ -145,6 +146,9 @@ def get_config(self):
         Returns:
             ProjectConfig. None if not set.
         """
+        self._condition.acquire()
```
oakbani: We should only be doing this for the first time. See Ruby.

optimizely/config_manager.py
```diff
@@ -187,10 +194,12 @@ def __init__(self,
         self.datafile_url = self.get_datafile_url(sdk_key, url,
                                                   url_template or enums.ConfigManager.DATAFILE_URL_TEMPLATE)
         self.set_update_interval(update_interval)
+        self.set_blocking_timeout(blocking_timeout)
         self.last_modified = None
         self._polling_thread = threading.Thread(target=self._run)
         self._polling_thread.setDaemon(True)
         self._polling_thread.start()
+        self._condition = threading.Condition()
```
oakbani: You will have to put this inside static config manager since you use it in it's method.

optimizely/config_manager.py
```diff
@@ -249,6 +258,31 @@ def set_update_interval(self, update_interval):
 
         self.update_interval = update_interval
 
+    def set_blocking_timeout(self, blocking_timeout):
+        """ Helper method to set time in seconds to block the config call until config has been initialized.
+
+        Args:
+          blocking_timeout: Time in seconds to block the config call.
+        """
+        if not blocking_timeout:
```
oakbani: nit. Check if it's None. Otherwise, this will be true when blocking_timeout is 0

optimizely/config_manager.py
```diff
@@ -249,6 +258,31 @@ def set_update_interval(self, update_interval):
...
+            blocking_timeout = enums.ConfigManager.DEFAULT_BLOCKING_TIMEOUT
+            self.logger.debug('Set config blocking timeout to default value {}.'.format(blocking_timeout))
```
oakbani: nit. See log in Ruby

optimizely/config_manager.py
```diff
@@ -249,6 +258,31 @@ def set_update_interval(self, update_interval):
...
+
+        if not isinstance(blocking_timeout, (int, float)):
```
oakbani: use numbers.Integral instead of int. This will cover long for Python 2 as well

optimizely/config_manager.py
```diff
@@ -249,6 +258,31 @@ def set_update_interval(self, update_interval):
...
+            raise optimizely_exceptions.InvalidInputException(
+                'Invalid blocking timeout "{}" provided.'.format(blocking_timeout)
+            )
+
+        # If blocking timeout is less than or equal to 0 then set it to default blocking timeout.
+        if blocking_timeout <= 0:
```
oakbani: Why haven't you enforced MIN and MAX limit as in Ruby SDK?

oakbani:
@rashidsp  Use threading.Event instead. https://docs.python.org/2.7/library/threading.html?highlight=condition%20acquire#event-objects

optimizely/config_manager.py
```diff
@@ -133,6 +135,7 @@ def _set_config(self, datafile):
             return
 
         self._config = config
+        self._configReadyEvent.set()
```
aliabbasrizvi: What is the effect of calling `set()` twice

optimizely/helpers/enums.py
```diff
@@ -38,6 +38,8 @@ class AudienceEvaluationLogs(object):
 
 class ConfigManager(object):
   DATAFILE_URL_TEMPLATE = 'https://cdn.optimizely.com/datafiles/{sdk_key}.json'
+  # Default time in seconds to block the 'get_config' method call until 'config' instance has been initialized.
+  DEFAULT_BLOCKING_TIMEOUT = 15
```
aliabbasrizvi: Let's make this 10 seconds.

optimizely/config_manager.py
oakbani: No effect. It will make the already 'true' internal flag to 'true' again. Once an event has been set, all the threads which wait for the events do not block until .clear() has been called on the event, which makes the internal flag false.

mikeng13:
Please include a link to the passing FSC tests with Event Batching enabled

PR #214, opened 2019-10-18
fix: avoid type error for Python 3

oakbani:
Summary
-------
- Fixes python 3 type error when flush interval passes as string
- Checks if self.executor is defined before calling method on it to avoid scenario where start_on_init is False and user tries to check if the processor is running or tries to stop it. 

Test plan
---------
Added unit test with flush interval passed as string

Issues
------
-  OASIS-5525

### DIFF
--- a/optimizely/event/event_processor.py
+++ b/optimizely/event/event_processor.py
@@ -12,6 +12,7 @@
 # limitations under the License.
 
 import abc
+import numbers
 import threading
 import time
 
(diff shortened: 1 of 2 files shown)


optimizely/event/event_processor.py
```diff
@@ -121,12 +123,18 @@ def _validate_intantiation_props(self, prop, prop_name):
       False if property name is batch_size and value is a floating point number.
       True otherwise.
     """
-    if (prop_name == 'batch_size' and not isinstance(prop, int)) or prop is None or prop <= 0 or \
-      not validator.is_finite_number(prop):
+    is_valid = True
+
+    if prop is None or not validator.is_finite_number(prop) or prop <= 0:
```
oakbani: The fix part is that **not validator.is_finite_number(prop)** comes before **prop<=0**

Mat001:
@oakbani I haven't tried testing Python 2. Can you pls check if Py 2 throws that error too. Let me know or Sohail please. thank you.( @msohailhussain )

msohailhussain:
> @oakbani I haven't tried testing Python 2. Can you pls check if Py 2 throws that error too. Let me know or Sohail please. thank you.( @msohailhussain )

No this issue was not in python2

PR #216, opened 2019-10-24
fix(Polling Manager): Fixing how config manager blocks. Miscellaneous other fixes as well.

aliabbasrizvi:
Summary
-------
- The polling config manager is supposed to wait until config is ready. Previously, the config will be attempted to be initialized with empty datafile. This was resolved by overriding up `_set_config` in `PollingConfigManager`.
- Config ready event had no business in `StaticConfigManager` so it has been moved from there.
- Some formatting and other miscellaneous fixes.

### DIFF
--- a/optimizely/event/event_processor.py
+++ b/optimizely/event/event_processor.py
@@ -34,7 +34,7 @@ class BaseEventProcessor(ABC):
   """ Class encapsulating event processing. Override with your own implementation. """
 
   @abc.abstractmethod
-  def process(user_event):
+  def process(self, user_event):
     """ Method to provide intermediary processing stage within event production.
     Args:
       user_event: UserEvent instance that needs to be processed and dispatched.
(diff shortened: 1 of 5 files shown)


optimizely/event/event_processor.py
```diff
@@ -110,13 +119,14 @@ def is_running(self):
     """ Property to check if consumer thread is alive or not. """
     return self.executor.isAlive() if self.executor else False
 
-  def _validate_intantiation_props(self, prop, prop_name):
+  def _validate_intantiation_props(self, prop, prop_name, default_value):
```
mjc1283: Nit: spelling
```suggestion
  def _validate_instantiation_props(self, prop, prop_name, default_value):
```

optimizely/event/event_processor.py
```diff
@@ -226,14 +235,17 @@ def _flush_queue(self):
...
       user_event: UserEvent Instance.
     """
     if not isinstance(user_event, UserEvent):
       self.logger.error('Provided event is in an invalid format.')
       return
 
-    self.logger.debug('Received user_event: ' + str(user_event))
+    self.logger.debug('Received event of type {} for user {}.'.format(
+      type(user_event).__name__, user_event.user_id)
+    )
```
mjc1283: Nit: Do we log this in other SDKs? This could be extremely chatty.

aliabbasrizvi: Yeah, at least in the Java SDK. It could be chatty, but I feel it is a useful message.

optimizely/event/event_processor.py
aliabbasrizvi: Nice catch. Will update.

optimizely/event/event_processor.py
```diff
@@ -86,20 +87,28 @@ def __init__(self,
...
     self.timeout_interval = timedelta(seconds=timeout_interval) \
-                              if self._validate_intantiation_props(timeout_interval, 'timeout_interval') \
-                              else self._DEFAULT_TIMEOUT_INTERVAL
-    self.notification_center = notification_center
+                              if self._validate_instantiation_props(timeout_interval,
+                                                                    'timeout_interval',
+                                                                    self._DEFAULT_TIMEOUT_INTERVAL) \
+                              else timedelta(self._DEFAULT_TIMEOUT_INTERVAL)
+
+    self.notification_center = notification_center or _notification_center.NotificationCenter(self.logger)
```
msohailhussain: What's the purpose of this line?

aliabbasrizvi: Helps us get rid of conditionals later on.

PR #221, opened 2019-11-13
Update setup.py

### LINKED ISSUE #220: Pip raises UnicodeDecodeError when installing on windows
`UnicodeDecodeError: 'charmap' codec can't decode byte 0x81 in position 7717: character maps to <undefined>
`

emielrv:
Fix windows encoding issue

Summary
-------

-  Fixed the encoding issue in windows of the changelog when pip installing

Issues
------

-  “Fixes #220 ”

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -20,7 +20,7 @@ with open(os.path.join(here, 'requirements', 'test.txt')) as _file:
 with open(os.path.join(here, 'README.md')) as _file:
   README = _file.read()
 
-with open(os.path.join(here, 'CHANGELOG.md')) as _file:
+with open(os.path.join(here, 'CHANGELOG.md'), encoding="utf-8") as _file:
   CHANGELOG = _file.read()
 
 about_text = 'Optimizely X Full Stack is A/B testing and feature management for product development teams. ' \


aliabbasrizvi:
@emielrv thank you for submitting a PR to address this. As a pre-requisite can you please sign our CLA mentioned in the contributing doc: https://github.com/optimizely/python-sdk/blob/master/CONTRIBUTING.md

emielrv:
I signed the CLA

setup.py
```diff
@@ -20,7 +20,7 @@
 with open(os.path.join(here, 'README.md')) as _file:
```
aliabbasrizvi: You may have to make the change here as well?

emielrv: Well, to be honest I don't know what would be a best practice here. I know only changing it here will solve the problem, since the rest doesn't have these special characters. But maybe it is a best practice to change it everywhere.

aliabbasrizvi: Looks like the change you made will not work with Python 2.7.

We can try updating the content in CHANGELOG itself.

aliabbasrizvi:
This was addressed in 3.3.1 release. Closing this PR.

PR #222, opened 2019-11-14
chore: Formatting code

aliabbasrizvi:


.flake8
```diff
@@ -0,0 +1,5 @@
+[flake8]
```
aliabbasrizvi: Moving to `.flake8` from `tox.ini`

PR #226, opened 2019-12-06
feat: get_optimizely_config API

oakbani:
Summary
-------
Adds support for get_optimizely_config public API. 

Test plan
---------
- Tested manually with FSC OptimizelyConfig tests. 
- Added simple Unit tests in test_optimizely.py

Issues
------
-  OASIS-5464

### DIFF
--- a/optimizely/optimizely_config.py
+++ b/optimizely/optimizely_config.py
@@ -76,7 +76,7 @@ class OptimizelyConfigService(object):
         self._create_lookup_maps()
 
     def get_config(self):
-        """ Returns instance of OptimizelyConfig
+        """ Gets instance of OptimizelyConfig
 
         Returns:
             Optimizely Config instance or None if OptimizelyConfigService is invalid.
--- a/optimizely/optimizely_config.py
+++ b/optimizely/optimizely_config.py
@@ -52,15 +52,25 @@ class OptimizelyVariable(object):
         self.value = value
 
 
-class OptimizelyConfigBuilder(object):
+class OptimizelyConfigService(object):
+    """ Class encapsulating methods to be used in creating instance of OptimizelyConfig. """
 
     def __init__(self, project_config):
+        """
+        Arguments:
+            project_config ProjectConfig
+        """
         self.experiments = project_config.experiments
         self.feature_flags = project_config.feature_flags
         self.groups = project_config.groups
         self.revision = project_config.revision
 
-    def build(self):
+    def get_optimizely_config(self):
+        """ Returns instance of OptimizelyConfig
+
+        Returns:
+            Optimizely Config instance.
+        """
         self._create_lookup_maps()
 
         experiments_key_map, experiments_id_map = self._get_experiments_maps()
@@ -69,6 +79,8 @@ class OptimizelyConfigBuilder(object):
         return OptimizelyConfig(self.revision, experiments_key_map, features_map)
 
     def _create_lookup_maps(self):
+        """ Creates lookup maps to avoid redundant iteration of config objects.  """
+
         self.exp_id_to_feature_map = {}
         for feature in self.feature_flags:
             for id in feature['experimentIds']:
@@ -90,6 +102,15 @@ class OptimizelyConfigBuilder(object):
             self.feature_key_variable_id_to_variable_map[feature['key']] = variables_id_map
 
     def _get_variables_map(self, variation, experiment):
+        """ Gets variables map for given variation and experiment.
+
+        Arguments:
+            variation dict
+            experiment dict
+
+        Returns:
+            dict - Map of variable key to OptimizelyVariable for the given variation.
+        """
         feature_flag = self.exp_id_to_feature_map.get(experiment['id'], None)
         if feature_flag is None:
             return {}
@@ -99,7 +120,7 @@ class OptimizelyConfigBuilder(object):
         variables_map = copy.deepcopy(self.feature_key_variable_key_to_variable_map[feature_flag['key']])
 
         # set variation specific variable value if any
-        if variation.get('featureEnabled', None):
+        if variation.get('featureEnabled'):
             for variable in variation.get('variables', []):
                 feature_variable = self.feature_key_variable_id_to_variable_map[feature_flag['key']][variable['id']]
                 variables_map[feature_variable.key].value = variable['value']
@@ -107,6 +128,14 @@ class OptimizelyConfigBuilder(object):
         return variables_map
 
     def _get_variations_map(self, experiment):
+        """ Gets variation map for the given experiment.
+
+        Arguments:
+            experiment dict
+
+        Returns:
+            dict -- Map of variation key to OptimizelyVariation.
+        """
         variations_map = {}
 
         for variation in experiment.get('variations', []):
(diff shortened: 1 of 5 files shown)


optimizely/optimizely_config.py
```diff
@@ -0,0 +1,213 @@
...
+        self.type = type
+        self.value = value
+
+
+class OptimizelyConfigService(object):
+    """ Class encapsulating methods to be used in creating instance of OptimizelyConfig. """
+
+    def __init__(self, project_config):
+        """
+        Arguments:
```
aliabbasrizvi: nit. Args.

optimizely/optimizely_config.py
```diff
@@ -0,0 +1,213 @@
...
+class OptimizelyVariation(object):
+    def __init__(self, id, key, feature_enabled, variables_map):
+        self.id = id
+        self.key = key
+        self.feature_enabled = feature_enabled
+        self.variables_map = variables_map
+
+
+class OptimizelyVariable(object):
+    def __init__(self, id, key, type, value):
```
aliabbasrizvi: type is a keyword and shadows a built in name. Would recommend naming this variable something else. May be `variable_type`

optimizely/optimizely_config.py
```diff
@@ -0,0 +1,213 @@
...
+        self.groups = project_config.groups
+        self.revision = project_config.revision
+
+    def get_optimizely_config(self):
+        """ Returns instance of OptimizelyConfig
+
+        Returns:
+            Optimizely Config instance.
+        """
+        self._create_lookup_maps()
```
aliabbasrizvi: This seems awkward that a get is calling a create.

optimizely/optimizely_config.py
```diff
@@ -0,0 +1,213 @@
...
+                variables_key_map[variable['key']] = opt_variable
+                variables_id_map[variable['id']] = opt_variable
+
+            self.feature_key_variable_key_to_variable_map[feature['key']] = variables_key_map
+            self.feature_key_variable_id_to_variable_map[feature['key']] = variables_id_map
+
+    def _get_variables_map(self, variation, experiment):
+        """ Gets variables map for given variation and experiment.
+
+        Arguments:
```
aliabbasrizvi: Args and not Arguments. Apply this feedback throughout this PR.

optimizely/optimizely_config.py
```diff
@@ -0,0 +1,213 @@
...
+        """
+        Arguments:
+            project_config ProjectConfig
+        """
+        self.experiments = project_config.experiments
+        self.feature_flags = project_config.feature_flags
+        self.groups = project_config.groups
+        self.revision = project_config.revision
+
+    def get_optimizely_config(self):
```
aliabbasrizvi: nit. Not sure if the naming has to be `get_optimizely_config`, but `OptimizelyConfigService.get_optimizely_config` has too much redundant information in it.

Something as simple as `get_config` may suffice here. cc @jaeopt

tests/test_optimizely.py
```diff
@@ -3911,6 +3912,39 @@ def test_get_feature_variable_returns__default_value__complex_audience_match(sel
...
+            'Invalid config. Optimizely instance is not valid. ' 'Failing "get_optimizely_config".'
+        )
+
+    def test_get_optimizely_config_returns_instance_of_optimizely_config(self):
+        """ Test that get_optimizely_config returns an instance of OptimizelyConfig. """
+
+        opt_obj = optimizely.Optimizely(json.dumps(self.config_dict_with_features))
+        opt_config = opt_obj.get_optimizely_config()
+        self.assertIsInstance(opt_config, optimizely_config.OptimizelyConfig)
+
```
jaeopt: Can we add tests for OptConfig contents validation as well?

oakbani: Just pushed.

oakbani:
@jaeopt  Have added unit tests. Please take a look.

tests/base.py
```diff
@@ -182,14 +182,12 @@ def setUp(self, config_dict='config_dict'):
                         {
                             'id': '122239',
                             'key': 'control',
-                            'featureEnabled': True,
-                            'variables': [{'id': '155551', 'value': '42.42'}],
+                            'variables': [],
```
oakbani: Removed these since this is not a feature experiment so shouldn't have these props.

optimizely/optimizely_config.py
```diff
@@ -0,0 +1,214 @@
...
+        self.id = id
+        self.key = key
+        self.type = variable_type
+        self.value = value
+
+
+class OptimizelyConfigService(object):
+    """ Class encapsulating methods to be used in creating instance of OptimizelyConfig. """
+
+    def __init__(self, project_config):
```
msohailhussain: do we need to check `instance` of `ProjectConfig`

optimizely/optimizely_config.py
```diff
@@ -0,0 +1,214 @@
...
+        features_map = self._get_features_map(experiments_id_map)
+
+        return OptimizelyConfig(self.revision, experiments_key_map, features_map)
+
+    def _create_lookup_maps(self):
+        """ Creates lookup maps to avoid redundant iteration of config objects.  """
+
+        self.exp_id_to_feature_map = {}
+        for feature in self.feature_flags:
+            for id in feature['experimentIds']:
```
msohailhussain: I would use instead of `id` use `experimentId`

optimizely/optimizely_config.py
```diff
@@ -0,0 +1,214 @@
...
+                self.exp_id_to_feature_map[id] = feature
+
+        self.feature_key_variable_key_to_variable_map = {}
+        self.feature_key_variable_id_to_variable_map = {}
+        for feature in self.feature_flags:
```
msohailhussain: am unable to find the difference between this `feature_flags` loop and at line 86. Can't we use one loop only?

optimizely/optimizely_config.py
oakbani: I don't expected to be used elsewhere. And in the main class, we validate project_config before using OptimizelyService. I can still validate if you so, but will have to keep a validity flag so that get_config returns nil and does not break.

msohailhussain: my suggestion is to validate but @aliabbasrizvi will you suggest for validation here?

aliabbasrizvi: Agree with @msohailhussain, we should validate here.

optimizely/optimizely_config.py
```diff
@@ -0,0 +1,225 @@
...
+
+        self.experiments = project_config.experiments
+        self.feature_flags = project_config.feature_flags
+        self.groups = project_config.groups
+        self.revision = project_config.revision
+
+        self._create_lookup_maps()
+
+    def get_config(self):
+        """ Returns instance of OptimizelyConfig
```
msohailhussain: gets

optimizely/optimizely_config.py
```diff
@@ -0,0 +1,225 @@
...
+                opt_variable = OptimizelyVariable(
+                    variable['id'], variable['key'], variable['type'], variable['defaultValue']
+                )
+                variables_key_map[variable['key']] = opt_variable
+                variables_id_map[variable['id']] = opt_variable
+
+            self.feature_key_variable_key_to_variable_map[feature['key']] = variables_key_map
+            self.feature_key_variable_id_to_variable_map[feature['key']] = variables_id_map
+
+    def _get_variables_map(self, variation, experiment):
```
aliabbasrizvi: nit. I personally prefer arranging this as `self, experiment, variation` to honor the parent child relationship between experiments and variations.

optimizely/optimizely_config.py
```diff
@@ -0,0 +1,225 @@
...
+        """ Gets variables map for given variation and experiment.
+
+        Args:
+            variation dict
```
aliabbasrizvi: This is very vague. Dict consisting of what?

zashraf1985:
@mikeng13 This is approved and i dont have rights to merge. Can you help.

PR #227, opened 2019-12-12
(fix): refactor batch_event_processor to reset deadline after it passes. 

thomaszurkan-optimizely:
Also, hang on queue with timeout at flush interval

Summary
-------
-  The deadline is not reset on the _run after the deadline has passed.  
-  get can hang with a timeout.  allow get to wait with timeout at flush interval

Test plan
---------
Current tests should pass

Issues
------

### DIFF
--- a/optimizely/event/event_processor.py
+++ b/optimizely/event/event_processor.py
@@ -184,6 +184,7 @@ class BatchEventProcessor(BaseEventProcessor):
                     self._flush_queue()
                     self.flushing_interval_deadline = self._get_time() + \
                         self._get_time(self.flush_interval.total_seconds())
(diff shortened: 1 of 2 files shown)


mikeng13:
Fix looks fine to me. Do we need additional unit tests for this fix?

optimizely/event/event_processor.py
```diff
@@ -159,9 +159,9 @@ def _get_time(self, _time=None):
       Integer time in seconds.
     """
         if _time is None:
-            return int(round(time.time()))
+            return time.time()
```
msohailhussain: Please change the description of the method.
Also Returns, valid time, can be float and integer. Need to revise documentation.

PR #228, opened 2019-12-14
(fix): add more debug logging. remove extra current_batch set

thomaszurkan-optimizely:
Summary
-------

-  Added more logging for why flush was initiated and flush size.
- Remove reset of current_batch that was not wrapped in a lock and was necessary after a call to flush_batch.

Test plan
---------
- Augmented test for flush once to include debug message for flushing with batch size.

### DIFF
--- a/optimizely/event/event_processor.py
+++ b/optimizely/event/event_processor.py
@@ -169,7 +169,7 @@ class BatchEventProcessor(BaseEventProcessor):
             self.logger.warning('BatchEventProcessor already started.')
             return
 
-        self.flushing_interval_deadline = self._get_time() + self._get_time(self.flush_interval.total_seconds())
+        self.flushing_interval_deadline = self._get_time() + self._get_time(self.flush_interval)
         self.executor = threading.Thread(target=self._run)
         self.executor.setDaemon(True)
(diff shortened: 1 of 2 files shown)


optimizely/event/event_processor.py
```diff
@@ -222,7 +222,7 @@ def flush(self):
 
     def _flush_queue(self):
         """ Flushes event_queue by dispatching events. """
-
+        self.logger.debug('Flushing the queue.')
```
aliabbasrizvi: Actually it will be better to look up if events are there and say nothing to flush if no events and then say flushing the queue if there are events. Also, perhaps number of user events flushed will be useful.

aliabbasrizvi:
Please add a better Summary / Test Plan. You have taken the template values right now.

optimizely/event/event_processor.py
```diff
@@ -267,6 +270,7 @@ def _add_to_batch(self, user_event):
       user_event: UserEvent Instance.
     """
         if self._should_split(user_event):
+            self.logger.debug('Flush on split.')
```
aliabbasrizvi: nit. Flushing and not flush.

PR #231, opened 2019-12-19
fix(batch_event_processor): small fix to start the processor if not started. By default the batch…

thomaszurkan-optimizely:
… processor thread is not started at init

Summary
-------

-  start the processor if not started. By default the batch processor thread is not started at init

The “why”, or other context.

Test plan
---------
- all unit tests should pass

### DIFF
--- a/optimizely/event/event_processor.py
+++ b/optimizely/event/event_processor.py
@@ -256,6 +256,12 @@ class BatchEventProcessor(BaseEventProcessor):
             'Received event of type {} for user {}.'.format(type(user_event).__name__, user_event.user_id)
         )
 
+        if not self.is_running:
+            try:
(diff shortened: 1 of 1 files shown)


optimizely/event/event_processor.py
```diff
@@ -256,6 +256,12 @@ def process(self, user_event):
             'Received event of type {} for user {}.'.format(type(user_event).__name__, user_event.user_id)
         )
 
+        if not self.is_running:
```
mikeng13: Do we have to worry about this processor having been stopped explicitly?

aliabbasrizvi:
@thomaszurkan-optimizely can you take a look at the comments here?

PR #234, opened 2020-01-13
feat: Cache OptimizelyConfig and set Pyyaml version

oakbani:
Summary
-------
Optimizely Config cached in Config Manager. OptimizelyConfig should only be created on datafile update. Otherwise, Config Manager should return a cached copy.

Set Pyyaml version to 5.2 since the latest version 5.3 only works with Python 3.5+. And we need to run tests with Python 3.4 on Travis

Test plan
---------
- Added unit tests

Issues
------
OASIS-5791

### DIFF
--- a/optimizely/config_manager.py
+++ b/optimizely/config_manager.py
@@ -1,4 +1,4 @@
-# Copyright 2019, Optimizely
+# Copyright 2019-2020, Optimizely
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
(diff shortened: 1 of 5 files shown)


oakbani:
> 
> 
> Looks good. One missing test case is to validate OptConfig data updated correctly after datafile changes. Can we add it?

Added. Please check.

oakbani:
@thomaszurkan-optimizely  @msohailhussain  I have fixed pyyaml version to 5.2 since the latest release only supports Python 2.7 or Python 3.5+. Have also updated the PR description.

PR #249, opened 2020-04-13
Adding version of python sdk to docs

yavorona:
Summary
-------

Added python sdk version to docs

Test plan
---------

None

Issues
------
OASIS-6103


docs/source/conf.py
```diff
@@ -14,11 +14,14 @@
 import sys
 sys.path.insert(0, os.path.abspath('../..'))
 
+from optimizely.version import __version__
```
aliabbasrizvi, line 17: This import should happen before the sys.path command

yavorona: Talked offline and agreed to disregard this comment.

PR #251, opened 2020-04-21
feat: Implement get_all_feature_variables and get_feature_variable_json

pawels-optimizely:
## Summary
- implement   get_all_feature_variables and get_feature_variable_json public APIs
- created and adjusted unit tests

### DIFF
--- a/optimizely/optimizely.py
+++ b/optimizely/optimizely.py
@@ -365,7 +365,7 @@ class Optimizely(object):
 
         self.notification_center.send_notifications(
             enums.NotificationTypes.DECISION,
-            enums.DecisionNotificationTypes.FEATURE_VARIABLES,
+            enums.DecisionNotificationTypes.ALL_FEATURE_VARIABLES,
             user_id,
             attributes or {},
             {
@@ -764,7 +764,7 @@ class Optimizely(object):
         )
 
     def get_feature_variable_json(self, feature_key, variable_key, user_id, attributes=None):
-        """ Returns value for a certain json variable attached to a feature.
+        """ Returns value for a certain JSON variable attached to a feature.
 
     Args:
       feature_key: Key of the feature whose variable's value is being accessed.
@@ -790,7 +790,7 @@ class Optimizely(object):
         )
 
     def get_all_feature_variables(self, feature_key, user_id, attributes=None):
-        """ Returns dictionary of all variables attached to a feature.
+        """ Returns dictionary of all variables and their corresponding values in the context of a feature.
 
     Args:
       feature_key: Key of the feature whose variable's value is being accessed.
@@ -798,7 +798,7 @@ class Optimizely(object):
       attributes: Dict representing user attributes.
 
     Returns:
-      Dictionary object of the variable. None if:
+      Dictionary mapping variable key to variable value. None if:
       - Feature key is invalid.
     """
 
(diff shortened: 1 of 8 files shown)


optimizely/optimizely.py
```diff
@@ -287,6 +287,96 @@ def _get_feature_variable_for_type(
...
+                'Returning default value for all variables of feature flag "%s".' % (user_id, feature_key)
+            )
+
+        all_variables = {}
+        for variable_key in feature_flag.variables:
+            variable = project_config.get_variable_for_feature(feature_key, variable_key)
+            variable_value = variable.defaultValue
+            if feature_enabled:
+                variable_value = project_config.get_variable_value_for_variation(variable, decision.variation)
+                self.logger.info(
```
aliabbasrizvi, line 347: This will be very noisy. Would recommend converting this to `debug` level.

optimizely/optimizely.py
```diff
@@ -287,6 +287,96 @@ def _get_feature_variable_for_type(
...
+            enums.DecisionNotificationTypes.FEATURE,
+            user_id,
+            attributes or {},
+            {
+                'feature_key': feature_key,
+                'feature_enabled': feature_enabled,
+                'source': decision.source,
+                'source_info': source_info,
+            },
+        )
```
aliabbasrizvi, line 377: This doesn't make sense. I think we should trigger multiple `feature-variable` notifications: https://docs.developers.optimizely.com/full-stack/docs/set-up-notification-listener-swift#section-set-up-a-decision-notification-listener

We do the same thing in `get_enabled_features` as well.

optimizely/project_config.py
```diff
@@ -106,6 +106,11 @@ def __init__(self, datafile, logger, error_handler):
                 )
 
         self.feature_key_map = self._generate_key_map(self.feature_flags, 'key', entities.FeatureFlag)
+        for feature in self.feature_key_map:
```
aliabbasrizvi, line 109: Add a note here explaining what is happening

tests/base.py
```diff
@@ -324,6 +328,8 @@ def setUp(self, config_dict='config_dict'):
                         {'id': '129', 'key': 'cost', 'defaultValue': '10.99', 'type': 'double'},
                         {'id': '130', 'key': 'count', 'defaultValue': '999', 'type': 'integer'},
                         {'id': '131', 'key': 'variable_without_usage', 'defaultValue': '45', 'type': 'integer'},
+                        {'id': '132', 'key': 'object', 'defaultValue': '{"test": 12}', 'type': 'string',
```
aliabbasrizvi, line 331: Can we add another variable here that has `type` json. That will be to confirm that it is parsed correctly.

optimizely/optimizely.py
pawels-optimizely: the idea is to return only one notification in get_all_feature_variables, and not multiple notifications.

optimizely/optimizely.py
```diff
@@ -287,6 +287,97 @@ def _get_feature_variable_for_type(
...
+        """ Helper method to determine value for all variables attached to a feature flag.
+
+    Args:
+      project_config: Instance of ProjectConfig.
+      feature_key: Key of the feature whose variable's value is being accessed.
+      user_id: ID for user.
+      attributes: Dict representing user attributes.
+
+    Returns:
+      Dictionary of all variables. None if:
```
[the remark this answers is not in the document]
pawels-optimizely: I think we still should return None - that way we are being consistent with _get_feature_variable_for_type()

PR #253, opened 2020-04-24
test: Remove benchmarking tests

oakbani:
Summary
-------
Benchmarking tests have become outdated. Removing them as discussed. 

Test plan
---------
All tests continue to pass.

Issues
------
- OASIS-6305


mikeng13:
Shouldn't we also stop them from running? I still see them in the status checks. Are you planning on doing that in a separate PR?

oakbani:
> Shouldn't we also stop them from running? I still see them in the status checks. Are you planning on doing that in a separate PR?

The one in the status is the fullstack-benchmark-suite. This PR removes the benchmarking python unit tests.

mikeng13:
> > Shouldn't we also stop them from running? I still see them in the status checks. Are you planning on doing that in a separate PR?
> 
> The one in the status is the fullstack-benchmark-suite. This PR removes the benchmarking python unit tests.

Yeah my bad. Lgtm

PR #268, opened 2020-05-27
fix(logs): Fixing log messages for Targeted Rollouts

aliabbasrizvi:
With this change we are introducing targeted rollout specific log messages. 

Since, rules right now do not have a "name", I am proposing to use rule's index in the log message.

### DIFF
--- a/optimizely/bucketer.py
+++ b/optimizely/bucketer.py
@@ -38,12 +38,12 @@ class Bucketer(object):
     def _generate_unsigned_hash_code_32_bit(self, bucketing_id):
         """ Helper method to retrieve hash code.
 
-    Args:
-      bucketing_id: ID for bucketing.
+        Args:
+            bucketing_id: ID for bucketing.
 
-    Returns:
-      Hash code which is a 32 bit unsigned integer.
-    """
+        Returns:
+            Hash code which is a 32 bit unsigned integer.
+        """
 
         # Adjusting MurmurHash code to be unsigned
         return mmh3.hash(bucketing_id, self.bucket_seed) & UNSIGNED_MAX_32_BIT_VALUE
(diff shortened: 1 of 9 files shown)


optimizely/bucketer.py
```diff
@@ -131,10 +132,6 @@ def bucket(self, project_config, experiment, user_id, bucketing_id):
         variation_id = self.find_bucket(project_config, bucketing_id, experiment.id, experiment.trafficAllocation)
         if variation_id:
             variation = project_config.get_variation_from_id(experiment.key, variation_id)
-            project_config.logger.info(
```
aliabbasrizvi, line 134: These log messages have been moved to the call site, akin to how log messages are generated for rollouts.

optimizely/bucketer.py
```diff
@@ -90,20 +90,21 @@ def find_bucket(self, project_config, bucketing_id, parent_id, traffic_allocatio
...
         """ For a given experiment and bucketing ID determines variation to be shown to user.
 
-    Args:
-      project_config: Instance of ProjectConfig.
-      experiment: Object representing the experiment for which user is to be bucketed.
-      user_id: ID for user.
-      bucketing_id: ID to be used for bucketing the user.
+        Args:
+            project_config: Instance of ProjectConfig.
+            experiment: Object representing the experiment or rollout rule in which user is to be bucketed.
```
mjc1283, line 95: Nit: maybe rename `experiment` to `experiment_or_rule`?

optimizely/helpers/audience.py
```diff
@@ -15,29 +15,39 @@
...
-  Args:
-    config: project_config.ProjectConfig object representing the project.
-    experiment: Object representing the experiment.
-    attributes: Dict representing user attributes which will be used in determining
-                if the audience conditions are met. If not provided, default to an empty dict.
-    logger: Provides a logger to send log messages to.
+    Args:
+        config: project_config.ProjectConfig object representing the project.
+        audience_conditions: Audience conditions corresponding to the experiment or rollout rule.
+        experiment_or_rollout_rule: String representing whether entity being evaluated is experiment or rollout rule.
```
mjc1283, line 33: I think this could be cleaner if:
- `audience_logs` was an argument, rather than `experiment_or_rollout_rule`
- The association between rollout rule/experiment and the corresponding logs was abstracted by the `Experiment` entity class, and internally represented by a dict, or class property/method (maybe make rollout rule its own entity class?)

optimizely/bucketer.py
aliabbasrizvi: Will change.

optimizely/helpers/audience.py
aliabbasrizvi: I will adopt the `audience_logs` change now.

> maybe make rollout rule its own entity class

I was thinking about that, but in light of more incoming product changes where this terminology will change a lot, I am hoping that we make the transition then. Let me know if that sounds ok.

optimizely/bucketer.py
aliabbasrizvi: I think I will keep it like this for now, given it is the `Experiment` entity.

PR #271, opened 2020-06-10
feat: Added support for Authenticated Datafiles

pthompson127:
Summary
-------
- Added support for Authenticated Datafiles
- Added new config manager class `(AuthDatafilePollingConfigManager)` that extends `PollingConfigManager` 
- Modify conditional logic in `optimizely.py` to determine the type of config manager that will be used depending on input parameters

Test Plan
-------
- Added tests for `AuthDatafilePollingConfigManager`
- Ran manual tests with a valid access token and corresponding sdk key

### DIFF
--- a/optimizely/config_manager.py
+++ b/optimizely/config_manager.py
@@ -398,7 +398,7 @@ class AuthDatafilePollingConfigManager(PollingConfigManager):
                              at which to request datafile and set ProjectConfig.
             blocking_timeout: Optional Time in seconds to block the get_config call until config object
                               has been initialized.
-            url: Optional string representing URL from where to fetch the authenticated datafile. 
+            url: Optional string representing URL from where to fetch the authenticated datafile.
                  If set it supersedes the sdk_key.
             url_template: Optional string template which in conjunction with sdk_key
                           determines URL from where to fetch the authenticated datafile.
--- a/optimizely/config_manager.py
+++ b/optimizely/config_manager.py
@@ -423,9 +423,9 @@ class AuthDatafilePollingConfigManager(PollingConfigManager):
         )
 
     def _set_access_token(self, access_token):
-        if access_token is None:
+        if not access_token:
             raise optimizely_exceptions.InvalidInputException(
-                'access_token cannot be None.')
+                'access_token cannot be empty or None.')
         self.access_token = access_token
 
     def fetch_datafile(self):
--- a/optimizely/config_manager.py
+++ b/optimizely/config_manager.py
@@ -383,13 +383,12 @@ class AuthDatafilePollingConfigManager(PollingConfigManager):
         *args,
         **kwargs
     ):
-        """ Initialize config manager. access_token must be set to be able to use.
-            One of sdk_key or url has to be set to be able to use.
+        """ Initialize config manager. One of sdk_key or url has to be set to be able to use.
 
         Args:
             access_token: String to be attached to the request header to fetch the authenticated datafile.
-            *args: Refer to arguments descriptions in PollingConfigManager
-            **kwargs: Refer to keyword arguments descriptions in PollingConfigManager
+            *args: Refer to arguments descriptions in PollingConfigManager.
+            **kwargs: Refer to keyword arguments descriptions in PollingConfigManager.
         """
         self._set_access_token(access_token)
         super(AuthDatafilePollingConfigManager, self).__init__(*args, **kwargs)
--- a/optimizely/config_manager.py
+++ b/optimizely/config_manager.py
@@ -371,7 +371,7 @@ class PollingConfigManager(StaticConfigManager):
 
 
 class AuthDatafilePollingConfigManager(PollingConfigManager):
-    """ Config manager that polls for authenticated datafile based on access token. """
+    """ Config manager that polls for authenticated datafile using access token. """
 
     def __init__(
         self,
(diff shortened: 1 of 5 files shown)


mikeng13:
Please wait for @aliabbasrizvi to take a look as well :)

aliabbasrizvi:
> Also, it would be easier to make a change in the code if users decided to instantiate PollingConfigManager themselves and switched to private datafiles instead.

I think the choice will be very explicit here.

optimizely/optimizely.py
```diff
@@ -40,6 +41,7 @@ def __init__(
         skip_json_validation=False,
         user_profile_service=None,
         sdk_key=None,
+        access_token=None,
```
aliabbasrizvi, line 44: This will break the SDK, since a user can choose not to use named arguments, but rather pass them in as positional arguments. This should be the last argument in the method.

Also, I am not sure if `access_token` needs to be introduced here. It is a concept in the config manager and one can simple invoke the `AuthDatafilePollingConfigManager` and pass that in. So this is probably not useful.

pawels-optimizely: good catch on the order of arguments, wondering how it passed the unit tests. 

IMO, adding access_token here  as the last optional argument does not hurt and even makes sense. Each config manager (Polling and Static) with default parameters is referenced in here, so not sure why AuthDatafilePollingConfigManager has to be the exception. If access_token was set on PollingConfigManager then this could be considered as an option just like poling interval, but AuthDatafilePollingConfigManager is a separate entity.

pthompson127: Done. Though, I think that `access_token` should still be included in the constructor in order to use `AuthDatafilePollingConfigManager` in the case where someone chooses to create an `optimizely` instance.

optimizely/config_manager.py
```diff
@@ -368,3 +368,76 @@ def start(self):
...
+            raise optimizely_exceptions.InvalidInputException(
+                'access_token cannot be empty or None.')
+        self.access_token = access_token
+
+    def fetch_datafile(self):
+        request_headers = {}
+        request_headers[enums.HTTPHeaders.AUTHORIZATION] = \
+            enums.ConfigManager.AUTHORIZATION_HEADER_DATA_TEMPLATE.format(access_token=self.access_token)
+
+        if self.last_modified:
```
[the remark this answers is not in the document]
pthompson127: 👍🏼

optimizely/config_manager.py
```diff
@@ -368,3 +368,76 @@ def start(self):
...
+        sdk_key=None,
+        datafile=None,
+        update_interval=None,
+        blocking_timeout=None,
+        url=None,
+        url_template=None,
+        logger=None,
+        error_handler=None,
+        notification_center=None,
+        skip_json_validation=False
```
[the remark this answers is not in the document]
pthompson127: Done. I've made the change, but because there is no logic in `PollingConfigManager` to handle the `url_template` on the condition where an access token is passed, I've added a helper method in `AuthDatafilePollingConfigManager` to put the appropriate `url_template` in `kwargs` (depending on the args that are passed in to `AuthDatafilePollingConfigManager` constructor) before passing it into the parent class's constructor.

optimizely/optimizely.py
```diff
@@ -89,14 +92,25 @@ def __init__(
...
+                    )
+                else:
+                    self.config_manager = PollingConfigManager(
+                        sdk_key=sdk_key,
+                        datafile=datafile,
+                        logger=self.logger,
+                        error_handler=self.error_handler,
+                        notification_center=self.notification_center,
+                        skip_json_validation=skip_json_validation,
+                    )
```
aliabbasrizvi, line 113: You can do something like:
```
config_manager_options = {
   'access_token':  access_token,
   'sdk_key': sdk_key,
   'notification_center': notification_center,
    ... <add all other common params>
    ... 
}
if access_token:
  config_manager_options['access_token'] = access_token
  self.config_manager = AuthDatafilePollingConfigManager(**config_manager_options)
else:
  self.config_manager = PollingConfigManager(**config_manager_options)
```

optimizely/config_manager.py
```diff
@@ -368,3 +370,47 @@ def start(self):
...
+
+    DATAFILE_URL_TEMPLATE = enums.ConfigManager.AUTHENTICATED_DATAFILE_URL_TEMPLATE
+
+    def __init__(
+        self,
+        access_token,
+        *args,
+        **kwargs
+    ):
+        """ Initialize config manager. access_token must be set to be able to use.
```
aliabbasrizvi, line 386: nit. It is implied that access_token is required. You can remove the comment.

tests/test_optimizely.py
```diff
@@ -252,6 +252,16 @@ def test_init__sdk_key_and_datafile(self):
 
         self.assertIs(type(opt_obj.config_manager), config_manager.PollingConfigManager)
 
+    def test_init__sdk_key_and_access_token(self):
+        """ Test that if both sdk_key and access_token is provided then AuthDatafilePollingConfigManager is used. """
+
+        with mock.patch('optimizely.config_manager.AuthDatafilePollingConfigManager._set_config'), mock.patch(
+            'threading.Thread.start'
+        ):
+            opt_obj = optimizely.Optimizely(access_token="test_access_token", sdk_key='test_sdk_key')
```
aliabbasrizvi, line 261: nit. You are interchangeably using single and double quotes. We follow single quotes.

optimizely/config_manager.py
```diff
@@ -368,3 +370,46 @@ def start(self):
         """ Start the config manager and the thread to periodically fetch datafile. """
         if not self.is_running:
             self._polling_thread.start()
+
+
+class AuthDatafilePollingConfigManager(PollingConfigManager):
```
aliabbasrizvi, line 375: Thinking more about this name. Should we just change this to `AuthenticatedDatafileConfigManager`? Polling is implied by it extending the `PollingConfigManager`.

pawels-optimizely: AuthDatafilePollingConfigManager is fine, and it explicitly mentions that this is a polling manager and not static manager.

PR #275, opened 2020-06-18
feat: add datafile accessor

pthompson127:
Summary
-------
- Add datafile accessor methods/functionalities

Testing
-------
- Ran manual tests

### DIFF
--- a/optimizely/optimizely_config.py
+++ b/optimizely/optimizely_config.py
@@ -24,10 +24,10 @@ class OptimizelyConfig(object):
         self.datafile = datafile
 
     def get_datafile(self):
-        """ Get the datafile associated with project_config.
+        """ Get the datafile associated with OptimizelyConfig.
 
         Returns:
-            A JSON string representation of the project datafile.
+            A JSON string representation of the environment's datafile.
         """
         return self.datafile
 
--- a/optimizely/optimizely_config.py
+++ b/optimizely/optimizely_config.py
@@ -17,10 +17,19 @@ from .project_config import ProjectConfig
 
 
 class OptimizelyConfig(object):
-    def __init__(self, revision, experiments_map, features_map):
+    def __init__(self, revision, experiments_map, features_map, datafile=None):
         self.revision = revision
         self.experiments_map = experiments_map
         self.features_map = features_map
+        self._datafile = datafile
+
+    def get_datafile(self):
+        """ Get the datafile associated with project_config.
+
+        Returns:
+            A JSON string representation of the project datafile.
+        """
+        return self._datafile
 
 
 class OptimizelyExperiment(object):
@@ -68,6 +77,7 @@ class OptimizelyConfigService(object):
             self.is_valid = False
             return
 
+        self._project_config = project_config
         self.experiments = project_config.experiments
         self.feature_flags = project_config.feature_flags
         self.groups = project_config.groups
@@ -88,7 +98,7 @@ class OptimizelyConfigService(object):
         experiments_key_map, experiments_id_map = self._get_experiments_maps()
         features_map = self._get_features_map(experiments_id_map)
 
-        return OptimizelyConfig(self.revision, experiments_key_map, features_map)
+        return OptimizelyConfig(self.revision, experiments_key_map, features_map, self._project_config.to_datafile())
 
     def _create_lookup_maps(self):
         """ Creates lookup maps to avoid redundant iteration of config objects.  """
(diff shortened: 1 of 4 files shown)


pthompson127:
Wanted to get thoughts on this implementation before proceeding (especially on `optimizely_config.py`). Constructor was changed due to the existence of `OptimizelyConfigService` and to enable `OptimizelyConfig` to get the datafile and match the same user interface described in Option 2.

optimizely/project_config.py
```diff
@@ -191,51 +192,60 @@ def get_typecast_value(self, value, type):
         else:
             return value
 
+    def to_datafile(self):
```
aliabbasrizvi, line 195: This should be called `get_datafile`

optimizely/optimizely_config.py
```diff
@@ -68,6 +77,7 @@ def __init__(self, project_config):
             self.is_valid = False
             return
 
+        self._datafile = project_config.to_datafile()
```
aliabbasrizvi, line 80: This should be `datafile = project_config.to_datafile()`. Doesn't have to be class variable.

optimizely/optimizely_config.py
```diff
@@ -17,10 +17,19 @@
...
 class OptimizelyConfig(object):
-    def __init__(self, revision, experiments_map, features_map):
+    def __init__(self, revision, experiments_map, features_map, datafile=None):
         self.revision = revision
         self.experiments_map = experiments_map
         self.features_map = features_map
+        self.datafile = datafile
+
+    def get_datafile(self):
+        """ Get the datafile associated with project_config.
```
aliabbasrizvi, line 27: Associated with `OptimizelyConfig`

optimizely/optimizely_config.py
```diff
@@ -17,10 +17,19 @@
...
+
+        Returns:
+            A JSON string representation of the project datafile.
```
aliabbasrizvi, line 30: `of the environment's datafile`

optimizely/project_config.py
pawels-optimizely: @mjc1283 and I think it should be to_datafile on project config and get_datafile on optimizely config

aliabbasrizvi: Why? `ProjectConfig` is an internal concept. And also, we are not doing any translation here.

optimizely/optimizely_config.py
pawels-optimizely: I think you can pass project_config.to_datafile() directly, otherwise I think it has to be class variable to access  it in class method

aliabbasrizvi: I see. Then in that case the call to `project_config.to_datafile` can be made in `get_config` itself right?

pawels-optimizely: correct, @pthompson127 please address it

optimizely/project_config.py
pawels-optimizely: @mjc1283, do you have a strong opinion ?

mjc1283: I like `to_datafile`. We aren't doing any translation, but if there is ever a `ProjectConfig` not backed by a datafile, we might do translation in that case. If @aliabbasrizvi strongly objects to `to_datafile`, I am OK with changing it to `get_datafile`.

optimizely/optimizely_config.py
mjc1283: I thought it was more in line with the surrounding code to only save reference to something derived from `project_config`, rather that save reference to the entire `project_config`.

pthompson127: Calling `project_config.to_datafile` in `get_config` requires a reference to `project_config`. Which one should I keep a reference to?: `project_config` or `datafile`?

pawels-optimizely: yes, I guess nevermind, we will leave datafile instead of project config, current implementation looks good

aliabbasrizvi: Current is fine.

optimizely/project_config.py
aliabbasrizvi: `to_datafile` is ok.

optimizely/project_config.py
```diff
@@ -191,51 +192,60 @@ def get_typecast_value(self, value, type):
...
+        """ Get the datafile.
```
aliabbasrizvi, line 196: Get the datafile corresponding to ProjectConfig.

PR #282, opened 2020-07-07
fix(logs): Updating feature variable evaluation logs

aliabbasrizvi:
Updating feature evaluation logs so that they apply to both rollouts and experiments

### DIFF
--- a/optimizely/project_config.py
+++ b/optimizely/project_config.py
@@ -477,16 +477,9 @@ class ProjectConfig(object):
 
         if variable_usage:
             variable_value = variable_usage.value
-            self.logger.info(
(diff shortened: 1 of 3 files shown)


optimizely/project_config.py
```diff
@@ -477,16 +477,9 @@ def get_variable_value_for_variation(self, variable, variation):
 
         if variable_usage:
             variable_value = variable_usage.value
-            self.logger.info(
-                'Value for variable "%s" for variation "%s" is "%s".' % (variable.key, variation.key, variable_value)
-            )
```
aliabbasrizvi, line 482: These messages combined with the ones in `optimizely.py` were redundant.

PR #283, opened 2020-07-09
fix: change datafile accessor feature to return a string representation of datafile

pthompson127:
Summary
---------
- Updated datafile to ensure that it is being stored and returned as a `string` type instead of previous `bytes` type

Test Plan
---------
- Ran manual tests

Note
-----
- Sohail mentioned that tests were failing in `fullstack_prod_suite` because the `bytes` datafile was not iterable, so I changed it to return a `string`

### DIFF
--- a/optimizely/project_config.py
+++ b/optimizely/project_config.py
@@ -40,7 +40,7 @@ class ProjectConfig(object):
         """
 
         config = json.loads(datafile)
-        self._datafile = config
+        self._datafile = str(datafile)
         self.logger = logger
         self.error_handler = error_handler
         self.version = config.get('version')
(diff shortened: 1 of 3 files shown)


optimizely/project_config.py
```diff
@@ -40,7 +40,7 @@ def __init__(self, datafile, logger, error_handler):
         """
 
         config = json.loads(datafile)
-        self._datafile = datafile
+        self._datafile = str(datafile)
```
aliabbasrizvi, line 43: You will have to specify encoding here if I understand correctly. It should be `utf-8`.

Python 2 by default does `ascii` and Python 3 does `utf-8`.

Mat001, line 43: This is per @mjc1283 's request. That's fine w me if we want to return a string object.

pthompson127:
Reran FSC tests with datafile accessor flag and passed. Merging.

PR #285, opened 2020-07-10
fix(exception-handling): Catch errors when requesting datafile

aliabbasrizvi:
Summary
-------

We were not catching all exceptions when requesting the datafile. The `HTTPError` class is only a single class of errors and we were not capturing all the other errors like timeout issues, proxy issues, etc.

The class of exceptions can be seen here: https://requests.readthedocs.io/en/master/_modules/requests/exceptions

Updating to catch `RequestException`. It is also something we were already doing correctly in event dispatching: https://github.com/optimizely/python-sdk/blob/master/optimizely/event_dispatcher.py#L42

Test plan
---------
- Tested timeout before and after the fix.
- Unit test added.

### DIFF
--- a/optimizely/config_manager.py
+++ b/optimizely/config_manager.py
@@ -325,7 +325,7 @@ class PollingConfigManager(StaticConfigManager):
         """
         try:
             response.raise_for_status()
-        except requests_exceptions.HTTPError as err:
+        except requests_exceptions.RequestException as err:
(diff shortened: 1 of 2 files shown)


aliabbasrizvi:
Tests had passed: https://travis-ci.org/github/optimizely/python-sdk/builds/706967427. Going to force merge.

mikeng13:
Go for it! Travis has been misbehaving lately...

PR #294, opened 2020-08-05
use named arguments in calls to public methods with more than one param

verygoodsoftwarenotvirus:
Summary
-------

- Calls made by `class Optimizely` to methods attached to `self` that are public and have more than one parameter are now made with named parameters.

Why
-------

In a private personal project, I implemented an Optimizely wrapper that inherits from the python-sdk client. I've implemented `activate` and `get_variation`, only I implemented `get_variation` with extra parameters relevant to my wrapper. Like so:

```python
def get_variation(self, experiment_key, other_info=None, excess_data=None, attributes=None):
```

My error in the above code is not having `attributes` be the third argument. When I called the wrapped `activate`, it calls `self.get_variation`, without named parameters. Because my code was wrong, but also because they are called with named parameters, values from `attributes` were being passed as `excess_data`, which caused unexpected behavior.

Test plan
---------
there are no syntactical changes that would be noticed by the interpreter in this changeset.

Issues
------
n/a

### DIFF
--- a/optimizely/optimizely.py
+++ b/optimizely/optimizely.py
@@ -411,7 +411,7 @@ class Optimizely(object):
             self.logger.error(enums.Errors.INVALID_PROJECT_CONFIG.format('activate'))
             return None
 
-        variation_key = self.get_variation(experiment_key, user_id, attributes)
+        variation_key = self.get_variation(experiment_key=experiment_key, user_id=user_id, attributes=attributes)
 
         if not variation_key:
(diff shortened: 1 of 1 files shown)


verygoodsoftwarenotvirus:
I don't know how to re-run the CI script, but my code didn't cause it to break. :shrug:

Tamara-Barum:
Closed as this is out of date/ old and no longer relevant.

PR #300, opened 2020-09-25
feat: support for send_flag_decisions

pawels-optimizely:
Summary
-------
- Add experiment/feature flag key, and  decision type to impression events.
- Send events for ALL decision types if send_flag_decisions is enabled.
- Add Metadata field to EventBatch.Decisions to capture metadata.flag_key, metadata.rule_key, metadata.rule_type, metadata.variation_key

https://optimizely.atlassian.net/browse/OASIS-7106

### DIFF
--- a/optimizely/event/event_factory.py
+++ b/optimizely/event/event_factory.py
@@ -90,7 +90,7 @@ class EventFactory(object):
 
         if isinstance(event, user_event.ImpressionEvent):
 
-            metadata = payload.Metadata(event.variation.key, event.flag_key, event.flag_type)
+            metadata = payload.Metadata(event.flag_key, event.flag_type, event.variation.key)
 
             decision = payload.Decision(event.experiment.layerId, event.experiment.id, event.variation.id, metadata)
 
--- a/optimizely/event/event_factory.py
+++ b/optimizely/event/event_factory.py
@@ -89,7 +89,7 @@ class EventFactory(object):
     """
 
         if isinstance(event, user_event.ImpressionEvent):
-            metadata = payload.Metadata(event.flag_key, event.flag_type, event.variation.key)
+            metadata = payload.Metadata(event.flag_key, event.rule_key, event.rule_type, event.variation.key)
             decision = payload.Decision(event.experiment.layerId, event.experiment.id, event.variation.id, metadata)
             snapshot_event = payload.SnapshotEvent(
                 event.experiment.layerId, event.uuid, cls.ACTIVATE_EVENT_KEY, event.timestamp,
--- a/optimizely/event/event_factory.py
+++ b/optimizely/event/event_factory.py
@@ -89,11 +89,8 @@ class EventFactory(object):
     """
 
         if isinstance(event, user_event.ImpressionEvent):
-
             metadata = payload.Metadata(event.flag_key, event.flag_type, event.variation.key)
-
             decision = payload.Decision(event.experiment.layerId, event.experiment.id, event.variation.id, metadata)
-
             snapshot_event = payload.SnapshotEvent(
                 event.experiment.layerId, event.uuid, cls.ACTIVATE_EVENT_KEY, event.timestamp,
             )
--- a/optimizely/event/event_factory.py
+++ b/optimizely/event/event_factory.py
@@ -89,7 +89,10 @@ class EventFactory(object):
     """
 
         if isinstance(event, user_event.ImpressionEvent):
-            decision = payload.Decision(event.experiment.layerId, event.experiment.id, event.variation.id,)
+
+            metadata = payload.Metadata(event.variation.key, event.flag_key, event.flag_type)
+
+            decision = payload.Decision(event.experiment.layerId, event.experiment.id, event.variation.id, metadata)
 
             snapshot_event = payload.SnapshotEvent(
                 event.experiment.layerId, event.uuid, cls.ACTIVATE_EVENT_KEY, event.timestamp,
--- a/optimizely/event/payload.py
+++ b/optimizely/event/payload.py
@@ -71,9 +71,10 @@ class Decision(object):
 class Metadata(object):
     """ Class respresenting Metadata. """
 
-    def __init__(self, flag_key, flag_type, variation_key):
+    def __init__(self, flag_key, rule_key, rule_type, variation_key):
         self.flag_key = flag_key
-        self.flag_type = flag_type
+        self.rule_key = rule_key
+        self.rule_type = rule_type
         self.variation_key = variation_key
 
 
--- a/optimizely/event/payload.py
+++ b/optimizely/event/payload.py
@@ -61,10 +61,19 @@ class EventBatch(object):
 class Decision(object):
     """ Class respresenting Decision. """
 
-    def __init__(self, campaign_id, experiment_id, variation_id):
+    def __init__(self, campaign_id, experiment_id, variation_id, metadata):
         self.campaign_id = campaign_id
         self.experiment_id = experiment_id
         self.variation_id = variation_id
+        self.metadata = metadata
+
+class Metadata(object):
+    """ Class respresenting Metadata. """
+
+    def __init__(self, variation_key, flag_key, flag_type):
+        self.variation_key = variation_key
+        self.flag_key = flag_key
+        self.flag_type = flag_type
 
 
 class Snapshot(object):
--- a/optimizely/event/payload.py
+++ b/optimizely/event/payload.py
@@ -67,6 +67,7 @@ class Decision(object):
         self.variation_id = variation_id
         self.metadata = metadata
 
+
 class Metadata(object):
     """ Class respresenting Metadata. """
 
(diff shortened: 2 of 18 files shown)


optimizely/event/payload.py
```diff
@@ -61,10 +61,20 @@ def get_event_params(self):
...
 
-    def __init__(self, campaign_id, experiment_id, variation_id):
+    def __init__(self, campaign_id, experiment_id, variation_id, metadata):
         self.campaign_id = campaign_id
         self.experiment_id = experiment_id
         self.variation_id = variation_id
+        self.metadata = metadata
+
+
+class Metadata(object):
```
aliabbasrizvi, line 71: nit. Would recommend making this a more descriptive name like `EventMetadata`

optimizely/event/user_event_factory.py
```diff
@@ -20,7 +20,7 @@ class UserEventFactory(object):
 
     @classmethod
     def create_impression_event(
-        cls, project_config, activated_experiment, variation_id, user_id, user_attributes,
+        cls, project_config, activated_experiment, variation_id, user_id, user_attributes, flag_key, flag_type
```
aliabbasrizvi, line 23: This new field is not documented. Also, would be good to rethink order of these params.

optimizely/optimizely.py
```diff
@@ -160,7 +160,7 @@ def _validate_user_inputs(self, attributes=None, event_tags=None):
 
         return True
 
-    def _send_impression_event(self, project_config, experiment, variation, user_id, attributes):
+    def _send_impression_event(self, project_config, experiment, variation, user_id, attributes, flag_key, flag_type):
```
aliabbasrizvi, line 163: Missing documentation for new fields. Also, may be good to rethink order of these parameters since it is an internal method.

optimizely/project_config.py
```diff
@@ -514,6 +515,15 @@ def get_anonymize_ip_value(self):
...
 
+    def get_send_flag_decisions_value(self):
+        """ Gets the Send Flag Decisions value.
+
+        Returns:
+            A boolean value that indicates if we should send flag decisions.
+        """
+
+        return self.send_flag_decisions
+
```
aliabbasrizvi, line 526: Can you add some tests in this file like we have for bot filtering: https://github.com/optimizely/python-sdk/blob/master/tests/test_config.py#L602-L614

aliabbasrizvi (changes requested):

optimizely/event/payload.py
pawels-optimizely: it is metadata in TDD and go-sdk implementation, so I just followed that

aliabbasrizvi: I think it is too generic.

optimizely/event/user_event_factory.py
```diff
@@ -36,10 +39,13 @@ def create_impression_event(
...
-        if not activated_experiment:
+        if not activated_experiment and flag_type is not enums.DecisionSources.ROLLOUT:
             return None
 
-        experiment_key = activated_experiment.key
+        if activated_experiment:
+            experiment_key = activated_experiment.key
+        else:
+            experiment_key = None
         variation = project_config.get_variation_from_id(experiment_key, variation_id)
```
aliabbasrizvi, line 49: How will this succeed if `experiment_key` is `None`? It will generate errors logs based on: https://github.com/optimizely/python-sdk/blob/master/optimizely/project_config.py#L340-L364.

optimizely/optimizely.py
```diff
@@ -160,19 +160,21 @@ def _validate_user_inputs(self, attributes=None, event_tags=None):
...
-    def _send_impression_event(self, project_config, experiment, variation, user_id, attributes):
+    def _send_impression_event(self, project_config, experiment, variation, flag_key, flag_type, user_id, attributes):
         """ Helper method to send impression event.
 
     Args:
       project_config: Instance of ProjectConfig.
       experiment: Experiment for which impression event is being sent.
       variation: Variation picked for user for the given experiment.
+      flag_key: key for a feature flag or experiment
+      flag_type: type for the source
```
aliabbasrizvi, line 171: Why are we not doing a metadata object vs passing in so many things around?

aliabbasrizvi (changes requested):

optimizely/event/payload.py
pawels-optimizely: if we are sending an event, I don't think there is a need to have EventMetadata, it already implies that the metadata belongs to an event.

optimizely/optimizely.py
pawels-optimizely: we only pass flag_key and flag_type around (2 things). metadata contains variation_key, flag_key,  and flag_type. 
here we pass variation and then we get id and key out of that. I think it's better to continue passing the things the way they are now

optimizely/optimizely.py
```diff
@@ -422,7 +424,8 @@ def activate(self, experiment_key, user_id, attributes=None):
 
         # Create and dispatch impression event
         self.logger.info('Activating user "%s" in experiment "%s".' % (user_id, experiment.key))
-        self._send_impression_event(project_config, experiment, variation, user_id, attributes)
+        self._send_impression_event(project_config, experiment, variation, experiment.key,
+                                    "experiment", user_id, attributes)
```
zashraf1985, line 428: @pawels-optimizely Should we add `experiment` to the `DecisionSources` enum instead of using hardcoded string?

pawels-optimizely: we can, go-sdk is using hardcoded value too, and since it's used only once I don't really care.

optimizely/event/event_factory.py
```diff
@@ -89,7 +89,10 @@ def _create_visitor(cls, event, logger):
     """
 
         if isinstance(event, user_event.ImpressionEvent):
-            decision = payload.Decision(event.experiment.layerId, event.experiment.id, event.variation.id,)
+
```
aliabbasrizvi, line 92: nit. Can remove this extra line.

optimizely/event/event_factory.py
```diff
@@ -89,7 +89,10 @@ def _create_visitor(cls, event, logger):
...
+            metadata = payload.Metadata(event.flag_key, event.flag_type, event.variation.key)
+
```
aliabbasrizvi, line 94: And this one as well.

optimizely/event/user_event_factory.py
```diff
@@ -36,12 +39,18 @@ def create_impression_event(
...
 
-        experiment_key = activated_experiment.key
-        variation = project_config.get_variation_from_id(experiment_key, variation_id)
+        if activated_experiment:
+            experiment_key = activated_experiment.key
+        else:
+            experiment_key = None
 
+        if variation_id:
+            variation = project_config.get_variation_from_id(experiment_key, variation_id)
```
aliabbasrizvi, line 51: I think I had mentioned previously. There will be an error message if experiment key is None. What is being done to counter that?

optimizely/optimizely.py
aliabbasrizvi: I think we should add to `DecisionSources` and use the value from there.

aliabbasrizvi (changes requested):

aliabbasrizvi (approved):

optimizely/event/user_event_factory.py
```diff
@@ -21,16 +21,17 @@ class UserEventFactory(object):
...
 
     Args:
       project_config: Instance of ProjectConfig.
       experiment: Experiment for which impression needs to be recorded.
       variation_id: ID for variation which would be presented to user.
-      flag_key: key for a feature flag or experiment
-      flag_type: type for the source
+      flag_key: key for a feature flag
+      rule_key: key for an experiment
+      rule_type: type for the source
```
aliabbasrizvi, line 34: nit. Periods at the end.

optimizely/optimizely.py
```diff
@@ -160,21 +160,23 @@ def _validate_user_inputs(self, attributes=None, event_tags=None):
...
 
     Args:
       project_config: Instance of ProjectConfig.
       experiment: Experiment for which impression event is being sent.
       variation: Variation picked for user for the given experiment.
-      flag_key: key for a feature flag or experiment
-      flag_type: type for the source
+      flag_key: key for a feature flag
+      rule_key: key for an experiment
+      rule_type: type for the source
```
aliabbasrizvi, line 173: Same as above.

optimizely/optimizely.py
```diff
@@ -424,7 +426,7 @@ def activate(self, experiment_key, user_id, attributes=None):
 
         # Create and dispatch impression event
         self.logger.info('Activating user "%s" in experiment "%s".' % (user_id, experiment.key))
-        self._send_impression_event(project_config, experiment, variation, experiment.key,
+        self._send_impression_event(project_config, experiment, variation, "", experiment.key,
```
aliabbasrizvi, line 429: nit. Use single quotes. That is the convention.

aliabbasrizvi (approved):

PR #303, opened 2020-10-14
fix: correcting metadata and decision payload

pawels-optimizely:
Summary
-------
- fixing metadata and decision payload

### DIFF
--- a/optimizely/event/event_factory.py
+++ b/optimizely/event/event_factory.py
@@ -89,10 +89,20 @@ class EventFactory(object):
     """
 
         if isinstance(event, user_event.ImpressionEvent):
-            metadata = payload.Metadata(event.flag_key, event.rule_key, event.rule_type, event.variation.key)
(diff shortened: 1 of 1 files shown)


optimizely/event/event_factory.py
```diff
@@ -89,10 +89,20 @@ def _create_visitor(cls, event, logger):
     """
 
         if isinstance(event, user_event.ImpressionEvent):
-            metadata = payload.Metadata(event.flag_key, event.rule_key, event.rule_type, event.variation.key)
-            decision = payload.Decision(event.experiment.layerId, event.experiment.id, event.variation.id, metadata)
+            experiment_layerId, experiment_id, variation_id, variation_key = '', '', '', ''
```
msohailhussain, line 92: `experiment_layer_id` or `experiment_layerid`

msohailhussain (approved):
lgtm

optimizely/event/event_factory.py
pawels-optimizely: yeah, I thought about it, but we already have  event.experiment.layerId

PR #305, opened 2020-11-09
feat: add user context

thomaszurkan-optimizely:
Summary
-------

- Added user context and create off of optimizely.  Added unit tests as well.

Test plan
---------
Unit tests added

### DIFF
--- a/optimizely/user_context.py
+++ b/optimizely/user_context.py
@@ -1,0 +1,84 @@
+#
+#    Copyright 2020, Optimizely and contributors
+#
+#    Licensed under the Apache License, Version 2.0 (the "License");
+#    you may not use this file except in compliance with the License.
+#    You may obtain a copy of the License at
+#
+#        http://www.apache.org/licenses/LICENSE-2.0
+#
+#    Unless required by applicable law or agreed to in writing, software
+#    distributed under the License is distributed on an "AS IS" BASIS,
+#    WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+#    See the License for the specific language governing permissions and
+#    limitations under the License.
+#
+
+
+class UserContext(object):
+    """
+    Representation of an Optimizely User Context using which APIs are to be called.
+    """
+
+    def __init__(self, optimizely_client, user_id, user_attributes=None):
+        """ Create an instance of the Optimizely User Context.
+
+        Args:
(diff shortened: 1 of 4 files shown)


tests/test_user_context.py
```diff
@@ -0,0 +1,34 @@
+# Copyright 2019, Optimizely
```
aliabbasrizvi, line 1: 2020

optimizely/user_context.py
```diff
@@ -0,0 +1,84 @@
+#
```
aliabbasrizvi, line 1: Omit this line.

optimizely/user_context.py
```diff
@@ -0,0 +1,83 @@
...
+        """
+        sets a attribute by key for this user context.
+        Args:
+          attribute_key: key to use for attribute
+          attribute_value: attribute value
+
+        Returns:
+        None
+        """
+        self.user_attributes[attribute_key] = attribute_value
```
jaeopt, line 48: Can we add a mutex control for attributes read/write?  
"setAttributes" and attribute -reading for decide can happen in other threads. See C#-sdk as a good reference (https://github.com/optimizely/csharp-sdk/pull/253)

tests/test_user_context.py
```diff
@@ -0,0 +1,34 @@
...
+        """
+        tests user context creating and attributes
+        """
+        uc = UserContext(self.optimizely, "test_user")
+        self.assertEqual(uc.user_attributes, {}, "should have created default empty")
+        self.assertEqual(uc.user_id, "test_user", "should have same user id")
+        uc.set_attribute("key", "value")
+        self.assertEqual(uc.user_attributes["key"], "value", "should have added attribute")
+        uc.set_attribute("key", "value2")
+        self.assertEqual(uc.user_attributes["key"], "value2", "should have new attribute")
```
jaeopt, line 34: Can we have more tests:
- construct with non-empty attributes
- setAttribute can override the init attributes
- attributes not changed once constructed when the caller copy is updated

jaeopt (changes requested):
Looks good. 
Should add mutex control for attributes. More tests suggested.

tests/test_user_context.py
oakbani: Done in 314

jaeopt:
Merged with #309

PR #312, opened 2021-01-12
feat: clone user context

ozayr-zaviar:
Summary
-------
-  `clone()` added in user_context to clone user_context and passed in `decide`, `decide_for_keys`and `decide_all` as argument

Test plan
---------
- All existing test cases should pass


oakbani (changes requested):
Looks good. Please write a unit test to verify that the UserContext object in decide API response is not the same object on which the decide api was called.

oakbani (approved):
Looks good.

oakbani:
Added changes in #314

PR #317, opened 2021-02-25
Fix PyPy unit test failure

Mat001:
Summary
-------

-  Fixing unit tests failing on PyPy/PyPy3
- bumped mmh3 library version to 3.0.0 as that helped eliminate some failures when testing locally

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,8 @@ python:
   - "3.4"
   - "3.5.5"
   - "3.6"
-# - "3.7" is handled in 'Test' job using xenial as Python 3.7 is not available for trusty.
-# - "3.8" is handled in 'Test' job using xenial as Python 3.8 is not available for trusty.
+  - "3.7"
+  - "3.8"
   - "pypy"
   - "pypy3"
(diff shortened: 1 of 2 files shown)


aliabbasrizvi:
Seems like a risky change given dropped support for Python 2.7, 3.3, 3.4: https://github.com/hajimes/mmh3/blob/master/README.md

Mat001:
> Seems like a risky change given dropped support for Python 2.7, 3.3, 3.4: https://github.com/hajimes/mmh3/blob/master/README.md

Good to see you, Ali!
Well, PyPy / PyPy3 unit tests have been failing because of mmh3.
Will not be rushing into breaking stuff for sure. Investigating atm ..

Mat001:
Addressed in #318 - note that failing dependency installations can be seen when expanding the view in TravisCI log console. A small triangle icon to the left of pip install -r requirements.txt... 

Also debating if worth downgrading cryptography lib (which caused filures) because that will cause other py versions tests to fail in the future. MAybe better to make a note in README for pypy/pypy3 to require a diff cryptography version

PR #319, opened 2021-03-03
chore: consolidate py versions

Mat001:
Summary
-------

The default Ubuntu version on TravisCI has now been Xenial (Ubuntu 16.04LTS) which supports Py 3.7 and 3.8.
Therefore the separation of unit tests into those for older Ubuntu Trusty (all but 3.7, 3.8) and Xenial (3.7, 3.8) is no longer needed.
Consolidating all Py versions under the same "python" section in .travis.tml file.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,8 +4,8 @@ python:
   - "3.4"
   - "3.5.5"
   - "3.6"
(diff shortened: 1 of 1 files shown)


Mat001:
Decided to abandon this PR. We're not introducing changes/improvements because they could introduce risk. We' are only a few devs at the moment, sticking to not breaking things, more maintenance, safe mode.

PR #320, opened 2021-03-04
ci: Install a compatible version of cryptography for pypy and pypy3 in travis config

msohailhussain:
Summary
-------
Unit tests on Python pypy and pypy3 have been failing consistently for quite some time. On detailed investigation, cryptography package is required by the requests[security] module. Although the latest cryptography version does support pypy, it doesn't build for some reason. 

Setting cryptography package to a static previous version revealed that Python 3.4 support has been dropped after cryptography 2.8.  

Therefore, instead of modifying requirements file, we install the correct working version of cryptography in before_install phase of pyp build. 

Test plan
---------
All checks continue to pass. 

Issues
------

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -64,6 +64,10 @@ jobs:
     - stage: 'Test'
       python: "pypy"
       before_install:
+        - pip install "cryptography>=1.3.4,<=3.1.1" # installing in before_install doesn't re-install a latest version of the same package in the next stage.
+    - stage: 'Test'
+      python: "pypy3"
+      before_install:
         - pip install "cryptography>=1.3.4,<=3.1.1"
     - stage: 'Test'
       dist: xenial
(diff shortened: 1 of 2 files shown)


.travis.yml
```diff
@@ -61,7 +61,14 @@ jobs:
         SDK=python
         SDK_BRANCH=$TRAVIS_PULL_REQUEST_BRANCH
         FULLSTACK_TEST_REPO=ProdTesting
-
+    - stage: 'Test'
+      python: "pypy"
+      before_install:
+        - pip install "cryptography>=1.3.4,<=3.1.1" # installing in before_install doesn't re-install a latest version of the same package in the next stage.
```
Mat001, line 67: Can you fix grammatical error: Change "a latest version of the same" to "the latest version of the same".

Mat001 (changes requested):

.travis.yml
oakbani: done

Mat001 (approved):
LGTM

PR #330, opened 2021-06-02
[OASIS-7552] fix: update requests library dependency

Mat001:
Summary
-------

- breaking down module requests[security] into its libraries because requests[security] will no longer be supported with requests version 2.26.
- I unpacked it into libraries: requests, pyOpenSSL, cryptography, idna. Looked for highest versions that work on all Py versions in the SDK (common denominator that works across all Py versions that we support in our SDK)

The “why”, or other context.

Test plan
---------
- FSC compatibility tests passing

Issues
------

-  https://optimizely.atlassian.net/browse/OASIS-7552

### DIFF
--- a/README.md
+++ b/README.md
@@ -224,3 +224,10 @@ would be:
 ### Contributing
 
 Please see [CONTRIBUTING](https://github.com/optimizely/python-sdk/blob/master/CONTRIBUTING.md).
+
+### Additional Code
+This software incorporates code from the following open source repos:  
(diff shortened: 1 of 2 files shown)


The-inside-man (approved):

jaeopt (approved):
It looks good to me, but not completely sure about implications.
We better get it confirmed with the python team.

Mat001:
> It looks good to me, but not completely sure about implications.
> We better get it confirmed with the python team.

good point

PR #335, opened 2021-06-07
Removing deprecated warnings

The-inside-man:
Summary
-------

        - assertRaisesRegexp -> assertRaisesRegex
        - assertEquals -> assertEqual
        - isAlive() -> is_alive()
        - Check added to base.py to confirm attribute assertRaisesRegex for backwards compatibility to Python2.7

Test plan
---------
 Compatibility Test

### DIFF
--- a/optimizely/event/event_processor.py
+++ b/optimizely/event/event_processor.py
@@ -1,4 +1,4 @@
-# Copyright 2019-2020 Optimizely
+# Copyright 2019-2021 Optimizely
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


keppel2:
Have you looked into preventing future warnings (lint)?

The-inside-man:
> Have you looked into preventing future warnings (lint)?

I was looking at the linter, but I wouldn't want to hide valid warnings. Since these deprecations are based on python and supporting libraries being upgraded the warnings are good in the sense it tells us developers to update our code base, but they shouldn't go untouched for too long. So keeping that in mind I don't want to block future warnings, just address them sooner maybe as they come so we are keeping our code base up to date as frequently as possible. For the most part updating to resolve these warnings are very little work and don't take much time. We just need to be concerned with backwards compatibility.

The-inside-man:
> Oh, I meant enforce a no-warning policy somehow. It would have to be situational, since some warnings are unavoidable.

ahh, ok , I see what you are saying! we can pass the --disable-warnings flag when using pytest or similar when building which would help for those who want to avoid. In the build scripts, this could be added to clean the output moving forward, we would then not see these warnings on the builds in CI. locally we would still see them unless we add that flag and same for anyone else using the SDK.

PR #339, opened 2021-06-24
Add attributes and events support to config

The-inside-man:
Summary
-------

- Attributes and Events support added to config
- Test cases added to test_optimizley_config

Test plan
---------

- Compatibility test

Issues
------

-  OASIS-7800

### DIFF
--- a/optimizely/optimizely_config.py
+++ b/optimizely/optimizely_config.py
@@ -25,7 +25,6 @@ class OptimizelyConfig(object):
         self.sdk_key = sdk_key
         self.environment_key = environment_key
 
-
     def get_datafile(self):
         """ Get the datafile associated with OptimizelyConfig.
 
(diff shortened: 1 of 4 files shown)


optimizely/optimizely_config.py
```diff
@@ -49,6 +52,22 @@ def get_environment_key(self):
...
+        returns:
+            A list of attributes.
+        """
+        return self.attributes
+
+    def get_events(self):
+        """ Get the events associated with OptimizelyConfig
+
+        returns:
+            A list of attributes.
```
keppel2, line 67: events

keppel2 (approved):

optimizely/optimizely_config.py
The-inside-man: Good catch! Thank you

Mat001 (approved):
LGTM

PR #348, opened 2021-07-14
chore(release): Preparing for 3.9.1 release

The-inside-man:

### DIFF
--- a/optimizely/event/user_event_factory.py
+++ b/optimizely/event/user_event_factory.py
@@ -45,12 +45,12 @@ class UserEventFactory(object):
         if not activated_experiment and rule_type is not enums.DecisionSources.ROLLOUT:
             return None
 
-        variation, experiment_key = None, None
+        variation, experiment_id = None, None
         if activated_experiment:
-            experiment_key = activated_experiment.key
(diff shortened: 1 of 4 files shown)


jaeopt (changes requested):
master has optConfigV2 fix that is not released yet. We need release this fix from v3.9.0

The-inside-man:
> master has optConfigV2 fix that is not released yet. We need release this fix from v3.9.0

I was told to do this PR, then create a new release branch from v3.9.0 and cherrypick this commit and the fix commit once ready. FSC has a few failures so need to make some changes first.

The-inside-man:
Once merged in I will create a new branch of release V3.9.0 commit and cherry pick this and the other fix commit into the new release branch which will be what we release for the Python SDK.

optimizely/event/user_event_factory.py
```diff
@@ -45,12 +45,12 @@ def create_impression_event(
         if not activated_experiment and rule_type is not enums.DecisionSources.ROLLOUT:
             return None
 
-        variation, experiment_key = None, None
+        variation, experiment_id = None, None
```
msohailhussain, line 48: update license header

PR #351, opened 2021-07-23
Delivery rules correction for proper rollout key

The-inside-man:
Summary
-------

-  Update the rollout key to use experiments instead of experiments_map
-  Update tests to have sample delivery rules

Test plan
---------
-  FSC

Issues
------

-  "OASIS-7812"

### DIFF
--- a/optimizely/optimizely_config.py
+++ b/optimizely/optimizely_config.py
@@ -325,7 +325,7 @@ class OptimizelyConfigService(object):
 
         for variation in experiment.get('variations', []):
             variables_map = self._get_variables_map(experiment, variation)
-            feature_enabled = variation.get('featureEnabled', None)
+            feature_enabled = variation.get('featureEnabled', False)
 
             optly_variation = OptimizelyVariation(
                 variation['id'], variation['key'], feature_enabled, variables_map
--- a/optimizely/optimizely_config.py
+++ b/optimizely/optimizely_config.py
@@ -443,7 +443,7 @@ class OptimizelyConfigService(object):
             for optly_audience in self.audiences:
                 audiences_map[optly_audience.id] = optly_audience.name
 
-            # Get the experiments_map for that rollout
+            # Get the experiments for that rollout
             experiments = rollout.get('experiments')
             if experiments:
                 for experiment in experiments:
--- a/optimizely/optimizely_config.py
+++ b/optimizely/optimizely_config.py
@@ -133,7 +133,7 @@ class OptimizelyConfigService(object):
             optly_audience = OptimizelyAudience(
                 typed_audience.get('id'),
                 typed_audience.get('name'),
-                typed_audience.get('conditions')
+                str(typed_audience.get('conditions'))
             )
             optly_typed_audiences.append(optly_audience)
             id_lookup_dict[typed_audience.get('id')] = typed_audience.get('id')
@@ -145,7 +145,7 @@ class OptimizelyConfigService(object):
                 optly_audience = OptimizelyAudience(
                     old_audience.get('id'),
                     old_audience.get('name'),
-                    old_audience.get('conditions')
+                    str(old_audience.get('conditions'))
                 )
                 optly_typed_audiences.append(optly_audience)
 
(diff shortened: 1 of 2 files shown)


tests/test_optimizely_config.py
```diff
@@ -677,7 +723,7 @@ def setUp(self):
                 },
                 'test_feature_in_experiment_and_rollout': {
                     'variables_map': {
-
+
```
The-inside-man, line 726: This is fine

The-inside-man:
@msohailhussain @jaeopt Python seems to pass FSC here without any errors which is interesting...

tests/test_optimizely_config.py
```diff
@@ -616,7 +616,53 @@ def setUp(self):
...
+                                    'variables_map': {}
+                                },
+                                '211229': {
+                                    'id': '211229',
+                                    'key': '211229',
+                                    'feature_enabled': False,
+                                    'variables_map': {}
+                                }
+                            },
+                            'audiences': ''
```
jaeopt, line 637: Do we have test cases to validate "audiences" in delivery rules (rollouts) are translated into OptimizelyExperiment? Otherwise, you can add valid audiences here.

jaeopt (approved):
One comment to clarify test cases. Other than that, LGTM

tests/test_optimizely_config.py
The-inside-man: Will add test cases to verify OptimizelyExperiment in delivery Rules.

keppel2:
To trigger continuous integration, I sometimes try to add or change a comment in .gitignore. I like to change the first line from '#' to '##' and also from '##' to '#'.

The-inside-man:
> To trigger continuous integration, I sometimes try to add or change a comment in .gitignore. I like to change the first line from '#' to '##' and also from '##' to '#'.

I fyou log into travis, you can actually just go to the build and trigger it there rather than pushing up new commits. I usually just do that, unless I have some functional item I am checking like None verse False.

optimizely/optimizely_config.py
```diff
@@ -436,7 +436,7 @@ def _get_delivery_rules(self, rollouts, rollout_id):
                 audiences_map[optly_audience.id] = optly_audience.name
 
             # Get the experiments_map for that rollout
```
keppel2, line 438: Comment does not match.

keppel2:
> > To trigger continuous integration, I sometimes try to add or change a comment in .gitignore. I like to change the first line from '#' to '##' and also from '##' to '#'.
> 
> If you log into travis, you can actually just go to the build and trigger it there rather than pushing up new commits. I usually just do that, unless I have some functional item I am checking like None verse False.

I'm not sure if that updates the statuses on GitHub. I think this is useful for failing builds on our public repos--we have had some intermittent issues in python.

optimizely/optimizely_config.py
The-inside-man: thanks!

The-inside-man:
> > > To trigger continuous integration, I sometimes try to add or change a comment in .gitignore. I like to change the first line from '#' to '##' and also from '##' to '#'.
> > 
> > 
> > If you log into travis, you can actually just go to the build and trigger it there rather than pushing up new commits. I usually just do that, unless I have some functional item I am checking like None verse False.
> 
> I'm not sure if that updates the statuses on GitHub. I think this is useful for failing builds on our public repos--we have had some intermittent issues in python.

Good call, I think you may be right on that. Rerunning from CI directly does not update the public build status.

optimizely/optimizely_config.py
```diff
@@ -133,7 +133,7 @@ def __init__(self, project_config):
             optly_audience = OptimizelyAudience(
```
keppel2, line 133: If you want, you can take a ref of typed_audiences using `[]`.

The-inside-man: Actually going to make it simpler and just use project_config.typed_audiences reference in the loop rather than assign something new. Was doing things a bit differently previously and never cleaned up. Thanks for spotting.

The-inside-man:
Passing FSC : https://travis-ci.com/github/optimizely/fullstack-sdk-compatibility-suite/jobs/529608777

The-inside-man:
All tests passing!

Mat001 (approved):
Looks ok to me.

Issue #354, opened 2021-07-29
BatchEventProcessor can try to deque event with negative timeout.


jkolenofferup:
The following code may ask for an item from the event_queue with a negative timeout interval.
```
class BatchEventProcessor(BaseEventProcessor):
    def _run(self):
        try:
            while True:
                if self._get_time() >= self.flushing_interval_deadline:
                    self._flush_batch()
                    self.flushing_interval_deadline = self._get_time() + \
                        self._get_time(self.flush_interval.total_seconds())
                    self.logger.debug('Flush interval deadline. Flushed batch.')
                try:
                    interval = self.flushing_interval_deadline - self._get_time()
                    item = self.event_queue.get(True, interval) ## interval can be negative
```
If the flushing_interval_deadline is between the first call of _get_time() and the second call of _get_time() then interval will be negative.

Proposed fix:
```
class BatchEventProcessor(BaseEventProcessor):
    def _run(self):
        try:
            while True:
                loop_time = self._get_time()  ## only call get_time once per loop iteration
                if loop_time >= self.flushing_interval_deadline:
                    self._flush_batch()
                    self.flushing_interval_deadline = self._get_time() + \
                        self._get_time(self.flush_interval.total_seconds())
                    self.logger.debug('Flush interval deadline. Flushed batch.')
                try:
                    interval = self.flushing_interval_deadline - loop_time
                    item = self.event_queue.get(True, interval) 
```

Mat001:
Thx @jkolenofferup . Can you provide a case where you were able to produce negative interval?

jkolenofferup:
It's sporadic. We have four jobs that each send about 500k events in a batch setting (batch size 100, flush time 10s).  Once we dealt with the event_queue dropping problem, we get about 30 to 40 error messages of the form "{event_processor.py:212} ERROR - Uncaught exception processing buffer. Error: 'timeout' must be a non-negative number@-@".
If you are trying to replicate in a unit test. 
- Set flushing_interval_time to T
- Let the first call to _get_time() return T
- Let the second call to _get_time() return T + epsilon
Anyways, making two requests for current time in a situation like this (expecting two calls to _get_time() to be constant) should be fixed.

Mat001:
@jkolenofferup I'm curious, do you still get the 30-40 error messages when you eliminate two _get_time() instances, and use loot_time instead?
Does your suggestion fix it?

jkolenofferup:
We were able to eliminate the error message by making the flush_interval
[...]

PR #356, opened 2021-08-04
[OASIS-7827] fix: make _get_time() value the same throughout the loop

Mat001:
Summary
-------

-  Running event processor would sometimes produce Buffer error with negative timeout interval for retrieving an event from the queue. Reported by a Github user (customer). See https://github.com/optimizely/python-sdk/issues/354 
- If the flushing_interval_deadline is between the first call of _get_time() and the second call of _get_time() then interval will be negative.

Test plan
---------
Unit tests for the _run() function where the change is made already extensively cover the _run() and event processor functionality. 
The change itself should prevent negative interval from occurring. To test that a larger load test would be required. 

Issues
------
-  https://optimizely.atlassian.net/browse/OASIS-7872

### DIFF
--- a/optimizely/event/event_processor.py
+++ b/optimizely/event/event_processor.py
@@ -180,14 +180,15 @@ class BatchEventProcessor(BaseEventProcessor):
     """
         try:
             while True:
-                if self._get_time() >= self.flushing_interval_deadline:
+                loop_time = self._get_time()
+                if loop_time >= self.flushing_interval_deadline:
                     self._flush_batch()
                     self.flushing_interval_deadline = self._get_time() + \
                         self._get_time(self.flush_interval.total_seconds())
                     self.logger.debug('Flush interval deadline. Flushed batch.')
 
                 try:
-                    interval = self.flushing_interval_deadline - self._get_time()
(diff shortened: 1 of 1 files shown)


optimizely/event/event_processor.py
```diff
@@ -180,14 +180,15 @@ def _run(self):
     """
         try:
             while True:
-                if self._get_time() >= self.flushing_interval_deadline:
+                loop_time = self._get_time()
+                if loop_time >= self.flushing_interval_deadline:
                     self._flush_batch()
                     self.flushing_interval_deadline = self._get_time() + \
```
The-inside-man, line 186: Should we change this ._get_time() to loop_time as well? - line 186

Mat001: Was thinking about that yes. Thx for the pointer. Will implement and test.

The-inside-man (approved):
All looks good to me. Noting that FSC will not pass till Config V2 is merged in. All other tests appear to pass which is good. Also noting to wait to merge till V2 Config is merged so we have a clean set of checks on this PR before merging.

PR #357, opened 2021-09-09
ISSUE-232---drop-mmh3-dependency

### LINKED ISSUE #232: Questions about dependencies
Hi, we're considering pulling in this SDK in several Python projects. Because we build those projects in pretty bare containers, the compiled dependencies here (`mmh3`, and `cryptography` from `requests[security]`) mean that we need to install some additional OS packages at build time. That's not a showstopper, but some questions came up:

* Is `requests[security]` needed? From https://github.com/psf/requests/pull/4825/files, it seems like it's probably not needed anymore, since fixes have been introduced in the standard library in Python 2.7.9 and 3.4.3.

* There is already fallback code in case `mmh3` isn't installed: https://github.com/optimizely/python-sdk/blob/d3ffe336de677e7dafff03253aa6c9adc49cf326/optimizely/bucketer.py#L16-L19, but the `mmh3` requirement doesn't seem to be optional. Could it be made optional?

* Do you have any details on the performance penalty of using the pure-Python Murmur3 implementation, as compared to the compiled version? I suspect that, in our use case (AWS Lambda, process per request), it may not be an issue.

GordonSo:
Summary
-------
What:
This PR is to remove the mmh3 package dependency thus c++  requirement on the build environment. See discussion: in ISSUE-232, https://github.com/optimizely/python-sdk/issues/232#issue-542968370

Why:
Windows OS developers are having to install c++ in order to use the package. It would be ideal to keep the code extendible at the import level but optional by choice for those who need it. 

Test plan
---------

Issues
------

-  “Fixes #232”

### DIFF
--- a/README.md
+++ b/README.md
@@ -35,13 +35,13 @@ To install:
 Note:  
 - If you are running the SDK with PyPy or PyPy3 and you are experiencing issues, install this cryptography package **first** and then optimizely-sdk package:  
 
-    pip install "cryptography>=1.3.4,<=3.1.1"
+    ```pip install "cryptography>=1.3.4,<=3.1.1"```
 
 
 - After v3.0.0 mmh3 is no longer a mandatory dependency to removed to remove the c++ build requirement.
 If you are experiencing (performance) problems, install this explicitly and it will continue to be honoured by code:
 
-     pip install "mmh3==2.5.1"
+     ```pip install "mmh3==2.5.1"```
 
 ### Feature Management Access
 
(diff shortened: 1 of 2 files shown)


GordonSo:
OK, I see the tests are coupled to mmh3 too, I can fix this. 
But before I do, can I get a confirmation this is the right path to go down, please?

msohailhussain:
Hi @GordonSo,
We are actively working on it, will get back to you once we are done with our research and decision.
@Mat001 FYI.

GordonSo:
Closing this as it will be fixed by:
https://github.com/optimizely/python-sdk/pull/362

PR #360, opened 2021-09-16
Add deprecation notes for OptimizleyConfigV2

The-inside-man:
Summary
-------

-  Deprecate experiments_map for OptimizelyFeature
-  Note added for experiments_map in OptimizelyConfig stating it is for legacy projects only

OptimizelyFeature experiments_map is deprecated, users shall use experiment_rules and delivery_rules.

Test plan
---------
- FSC

Issues
------
- N/A


jaeopt (approved):
Can you change the PR note. OptimizelyConfig experiments_map is not deprecated while that for OptimizelyFeature is. It's still valid for legacy projects. Other than that, LGTM

PR #362, opened 2021-10-21
enhancement: Remove mmh3 external dependency and use internal pymmh3 in place

The-inside-man:
Summary
-------
-  Removal of mmh3 external dependency and update imports to use internal pymmh3 instead

Testing
-------
- FSC

Issues
-------
- "OASIS-7995"

### DIFF
--- a/optimizely/bucketer.py
+++ b/optimizely/bucketer.py
@@ -13,10 +13,7 @@
 
 import math
 
-try:
-    import mmh3
-except ImportError:
-    from .lib import pymmh3 as mmh3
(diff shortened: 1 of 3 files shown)


The-inside-man:
This PR is in attention to: https://github.com/optimizely/python-sdk/issues/232

Mat001 (approved):
If all the tests pass and if FSC passes for older versions of Py then I think we're good.
LGTM

The-inside-man:
@msohailhussain Confirmed tests work on windows machine

Issue #165, opened 2019-03-08
Missing six package dependency from requirements.txt


asaschachar:
The Python SDK uses the package [six](https://github.com/optimizely/python-sdk/blob/fafad4cb53a4f6c3ab8e1ea0beede50e08bcf979/optimizely/decision_service.py#L15) but does not appear to require it in any of the [requirements](https://github.com/optimizely/python-sdk/blob/master/requirements/core.txt)

amilstead:
Looks like the _probable_ right version to use would be either `six>=1.4.1` (to match [`cryptography`](https://github.com/pyca/cryptography/blob/master/setup.py#L290)) or `six>=1.5.2`, (to match [`PyOpenSSL`](https://github.com/pyca/pyopenssl/blob/master/setup.py#L98)), mostly to be courteous to upstream deps. 

If not, you could always just use `>=1.12.0`, since that's the most recent version of six and satisfies existing dependencies.

From [pipdeptree](https://github.com/naiquevin/pipdeptree):

```
$ pipdeptree --reverse --packages optimizely-sdk,six
optimizely-sdk==3.1.0
six==1.12.0
  - cryptography==2.6.1 [requires: six>=1.4.1]
    - pyOpenSSL==19.0.0 [requires: cryptography>=2.3]
  - mock==1.3.0 [requires: six>=1.7]
  - pyOpenSSL==19.0.0 [requires: six>=1.5.2]
  - python-coveralls==2.7.0 [requires: six]
  - python-dateutil==2.7.5 [requires: six>=1.5]
    - botocore==1.12.62 [requires: python-dateutil>=2.1,<3.0.0]
      - awscli==1.16.72 [requires: botocore==1.12.62]
      - s3transfer==0.1.13 [requires: botocore>=1.3.0,<2.0.0]
        - awscli==1.16.72 [requires: s3transfer>=0.1.12,<0.2.0]
```

PR #363, opened 2021-11-09
fix: Add six dependency to requirements/core.txt

The-inside-man:
Summary
-------

-  Add six dependency as it was missing from requirements/core.txt

Note: This library is used throughout the core functionality of the SDK, however the build did not fail previously due to the six library being a transient dependency for jsonschema. Jsonschema is already in requirements/core.txt, however, we are now choosing to explicitly require the six dependency as we use this library explicitly in our SDK.

Test plan
---------
- FSC

Issues
------

-  https://github.com/optimizely/python-sdk/issues/165


msohailhussain (approved):
lgtm. per our discussion need to add description in the summary.

PR #364, opened 2021-11-09
Bump up PyYaml to recommended version to remediate vulnerability - DROP PYTHON 3.4 SUPPORT

The-inside-man:
Summary
-------

-  Removing PyYaml from SDK test.txt
-  Dropping  Python 3.4

Dependabot found critical vulnerability in this dependency and it is not needed in the SDK
Python 3.4 no longer supports required libraries to support new and existing features

Note: PyYaml was previously an explicit dependency, however PyYaml is also a transient dependency in python-coverals, which has already upgraded to the latest version of PyYaml. This version of PyYaml no longer supports Python version 3.4. For this reason we have chose to also drop Python 3.4, as we need to continue proper code coverage on our SDKs to ensure the highest quality of code.

Test plan
---------
- FSC

Issues
------

-  “OASIS-8054"

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 language: python 
 python:
   - "2.7"
-  - "3.4"
+#  - "3.4"
   - "3.5.5"
   - "3.6"
 # - "3.7" is handled in 'Test' job using xenial as Python 3.7 is not available for trusty.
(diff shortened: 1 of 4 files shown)


The-inside-man:
This is failing for Python 3.4 due to an environment issue it seems. Travis is still picking up the PyYaml requirement and trying to install, which it should not.

msohailhussain (approved):
lgtm

PR #367, opened 2021-12-13
fix: Forced variation not available in experiment

ozayr-zaviar:
Summary
-------

-  Crash fixed after logging syntax correction
-  Updated order for checking variation in impression event
- Unit test added for no active experiment in creating impression event
- Unit test fixed after changes

Test plan
---------
- All unit tests and FSC should pass.

### DIFF
--- a/optimizely/event/user_event_factory.py
+++ b/optimizely/event/user_event_factory.py
@@ -49,11 +49,12 @@ class UserEventFactory(object):
         if activated_experiment:
             experiment_id = activated_experiment.id
 
-        if variation_id and experiment_id:
-            variation = project_config.get_variation_from_id_by_experiment_id(experiment_id, variation_id)
-        # need this condition when we send events involving forced decisions
-        elif variation_id and flag_key:
+        if variation_id and flag_key:
             variation = project_config.get_flag_variation(flag_key, 'id', variation_id)
+        elif variation_id and experiment_id:
+            # need this condition when we send events involving forced decisions
+            variation = project_config.get_variation_from_id_by_experiment_id(experiment_id, variation_id)
+        
         event_context = user_event.EventContext(
             project_config.account_id, project_config.project_id, project_config.revision, project_config.anonymize_ip,
         )
--- a/optimizely/event/user_event_factory.py
+++ b/optimizely/event/user_event_factory.py
@@ -54,7 +54,7 @@ class UserEventFactory(object):
         if not variation and variation_id and experiment_id:
             # need this condition when we send events involving forced decisions
             variation = project_config.get_variation_from_id_by_experiment_id(experiment_id, variation_id)
-        
+
         event_context = user_event.EventContext(
             project_config.account_id, project_config.project_id, project_config.revision, project_config.anonymize_ip,
         )
--- a/optimizely/event/user_event_factory.py
+++ b/optimizely/event/user_event_factory.py
@@ -51,8 +51,8 @@ class UserEventFactory(object):
 
         if variation_id and flag_key:
             variation = project_config.get_flag_variation(flag_key, 'id', variation_id)
-        if not variation and variation_id and experiment_id:
-            # need this condition when we send events involving forced decisions
+        elif variation_id and experiment_id:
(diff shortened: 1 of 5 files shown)


dustin-sier (approved):
LGTM

optimizely/event/user_event_factory.py
```diff
@@ -49,11 +49,12 @@ def create_impression_event(
         if activated_experiment:
             experiment_id = activated_experiment.id
 
-        if variation_id and experiment_id:
-            variation = project_config.get_variation_from_id_by_experiment_id(experiment_id, variation_id)
-        # need this condition when we send events involving forced decisions
-        elif variation_id and flag_key:
+        if variation_id and flag_key:
             variation = project_config.get_flag_variation(flag_key, 'id', variation_id)
+        if not variation and variation_id and experiment_id:
```
jaeopt, line 54:
```suggestion
        elif variation_id and experiment_id:
```
We need to keep this to support legacy `activate` impression event.

optimizely/event/user_event_factory.py
```diff
@@ -49,11 +49,12 @@ def create_impression_event(
...
+            # need this condition when we send events involving forced decisions
```
jaeopt, line 55:
```suggestion
            # need this condition when we send events involving forced decisions (F-to-D or E-to-D with any ruleKey/variationKey combinations)
```
Also, Move this comment line to line 52

jaeopt: I'm also concerned about line 45-46, which looks like filtering any impression event request with None experiment but FEATURE-TEST source. A valid F-to-D will create such a request for impression event. 
Can you check if we have unit tests and FSC test covering this case - impression event fired for F-to-D?

jaeopt (changes requested):
Need changes and tests

msohailhussain (approved):
lgtm. please change the title of the PR.
Add proper comments and always recommend to add unit test for this case.

optimizely/event/user_event_factory.py
msohailhussain: This is a good & critical case, sure I will check it out.

tests/test_user_event_factory.py
```diff
@@ -70,6 +70,17 @@ def test_impression_event__with_attributes(self):
...
+    def test_impression_event__no_active_experiment(self):
+        project_config = self.project_config
+        user_id = 'test_user'
+
+        user_attributes = {'test_attribute': 'test_value', 'boolean_key': True}
+
+        impression_event = UserEventFactory.create_impression_event(
+            project_config, None, '111128', 'flag_key', 'rule_key', 'rule_type', True, user_id, user_attributes
+        )
+        self.assertIsNone(impression_event)
```
jaeopt, line 82: this should not be none. With FD, {None experiment} is a valid case, for which we should forward the impression event.

optimizely/event/user_event_factory.py
```diff
@@ -51,10 +51,10 @@ def create_impression_event(
 
         if variation_id and flag_key:
             variation = project_config.get_flag_variation(flag_key, 'id', variation_id)
-        if not variation and variation_id and experiment_id:
-            # need this condition when we send events involving forced decisions
+        elif variation_id and experiment_id:
+            # need this condition when we send events involving forced decisions (F-to-D or E-to-D with any ruleKey/variationKey combinations)
```
jaeopt, line 55: Move this comment line to the top of line 52.

jaeopt: Can we have a unit test covering this bug?

jaeopt (changes requested):
A couple of changes suggested.

optimizely/event/user_event_factory.py
msohailhussain: We checked it out, and this line won't make it null. 
https://github.com/optimizely/python-sdk/blob/master/optimizely/optimizely.py#L201

The-inside-man (approved):
lgtm!

jaeopt (approved):
LGTM

jaeopt (approved):
LGTM

PR #371, opened 2022-01-05
[fix] : [DO NOT REVIEW] Un-pin dependencies jsonschema and pyrsistent

The-inside-man:
Summary
-------

-  Unpinning dependencies jsonschema and pyrsistent so customer can use later packages.

Test plan
---------
- FSC

Issues
------


Mat001:
Too old and no longer relevant. Closing.

Issue #370, opened 2022-01-04
Dependency version pinning


gwerbin:
I am trying to use the Optimizely SDK, but my application also requires `jsonschema >= 4.0.1`, because it includes support for the newer JSONSchema specification drafts. Optimizely SDK requires exactly `jsonschema = 3.2.0`, so I am not able to use them together.

Is it possible to un-pin the version of `jsonschema` and `pyrsistent`?

Pinning dependencies in libraries (as opposed to "applications") is very unusual, and puts a lot of restrictions on users.

Normally one would expect that a minimum version is set, of course, but not an exact pinned version! 

And in this case, it also seems like the version pin isn't even necessary, since the [last version bump](https://github.com/optimizely/python-sdk/pull/239) didn't seem to require any code changes.

For example, it is effectively impossible to use Optimizely SDK and support OpenAPI Specification 3.1 in the same application. This is because OAS 3.1 specifies a JSONSchema draft that is not available in `jsonschema 3.2.0`. Practically, most schemas can be parsed with Draft 7, which _is_ available in 3.2.0, so you can mostly work around the problem. But it was surprising to find this dependency conflict at all.

Is there a compelling reason to keep dependencies pinned to exact patch versions in an SDK/library? It would make this much mroe user-friendly if these constraints were relaxed.

I see that dependency versions were discussed [in the past](https://github.com/optimizely/python-sdk/issues/232#issuecomment-597953558), but I didn't see any discussion of un-pinning.

Mat001:
@gwerbin Thanks for your request.
The reason we're capping the two versions is because we're still supporting older Py versions that don't work on newer versions of jsonschema and pyrsistent.

When I tested unpinned versions of jsonschema dn pyrsistent Py 2.7, 3.5 and PyPy2 fail, unable to find a suitable jsonschema version.

We've been debating internally removing support for some older Py versions (2, 3.5, pypy2). If we go ahead that would allow us to perform the unpinning.
I need to find out if there are and customer dependencies on those older versions and in general bring it up for discussion in the team.

Besides that I'm trying to find a version range that would allow jsonschema 4.0.1 (at least) while having older Py versions pass.

gwerbin:
Thanks for the reply and explanation, @Mat001.

Maybe the version should be a `<=` constraint and not a `==`? At least then you aren't pinning to a precise patch version, even if it doesn't solve the particular problem I stated in my original post.

Mat001:
[...]

PR #373, opened 2022-01-06
chore: remove old py versions

Mat001:
Summary
-------

-  Old python versions for Py SDK are causing security risks and cap using more up to date library versions. This creates issues for customers. 
For example we removed Py v3.4 already because PyYaml library in Py 3.4 can only be less than v5.4 which doesn't have a necessary security patch that versions 5.4 and greater do.
Another example is customer issue [issue link here] that required more up to date jsonschema lib version. But we could not provide that up to date version because Py 2.7 and 3.5 were capping it. See issue [370](https://github.com/optimizely/python-sdk/issues/370)
These are all issues related to supporting older Py versions that only work with libraries up certain version number. So we get a discrepancy between which library versions old Py and new Py support. Then we decided to remove support for older py version in order to free up support for up to date libraries. 
- We're removing Py 2, 3.5, 3.6, PyPy2. These are fairly old, none of them are officially supported in most libraries.

Test plan
---------
Unit tests, FSC

Issues
------

-  [OASIS-8124](https://optimizely.atlassian.net/browse/OASIS-8124)

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,5 @@
-language: python 
+dist: focal
+language: python
 python:
 #  - "2.7"
 #  - "3.4"
@@ -72,16 +73,12 @@ jobs:
       before_install:
         - pip install "cryptography>=1.3.4,<=3.1.1"
     - stage: 'Test'
-      dist: focal
       python: "3.7"
     - stage: 'Test'
-      dist: focal
       python: "3.8"
     - stage: 'Test'
-      dist: focal
       python: "3.9"
     - stage: 'Test'
-      dist: focal
       python: "3.10"
 
     - stage: 'Source Clear'
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,7 @@
 dist: focal
 language: python
 python:
-  - "pypy3"
+  - "pypy3.7-7.3.5"
   - "3.7"
   - "3.8"
   - "3.9"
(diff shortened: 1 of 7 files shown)


.travis.yml
```diff
@@ -35,10 +34,10 @@ jobs:
 
     - stage: 'Linting'
       language: python
-      python: "2.7"
+      python: "3.9"
       # flake8 version should be same as the version in requirements/test.txt
       # to avoid lint errors on CI
-      install: "pip install flake8==3.6.0" 
+      install: "pip install flake8>=3.6.0"
```
The-inside-man, line 40: Id say we should push this to our base supported version of 3.7 also

.travis.yml
```diff
@@ -61,20 +60,18 @@ jobs:
...
         SDK_BRANCH=$TRAVIS_PULL_REQUEST_BRANCH
         FULLSTACK_TEST_REPO=ProdTesting
-    - stage: 'Test'
-      python: "pypy"
-      before_install:
-        - pip install "cryptography>=1.3.4,<=3.1.1" # installing in before_install doesn't re-install the latest version of the same package in the next stage.
     - stage: 'Test'
       python: "pypy3"
       before_install:
         - pip install "cryptography>=1.3.4,<=3.1.1"
```
The-inside-man, line 66: Do we still need to be less than here with the bumped versions?

requirements/core.txt
```diff
@@ -1,7 +1,7 @@
-jsonschema==3.2.0
-pyrsistent==0.16.0
+jsonschema>=3.2.0
+pyrsistent>=0.16.0
 requests>=2.21
 pyOpenSSL>=19.1.0
-cryptography>=2.8.0
+cryptography>=2.8.0,<=3.1.1
```
The-inside-man, line 5: Same comment here as above.

The-inside-man (commented):
Couple suggestions.

.travis.yml
Mat001: Yeah, my bad, removed it now. Let's see if pypy3 tests compile

The-inside-man (approved):
LGTM!

PR #378, opened 2022-02-18
feat: BatchEventProcessor as Default Event Processor

ozayr-zaviar:
Summary
-------
- ForwardingEventProcessor sends calls in a synchronous manner so to reduce the time it is replaced with BatchEventProcessor which sends calls in an asynchronous manner.

Test plan
---------
- All unit tests and FSC tests should pass.


msohailhussain (approved):
lgtm

andrewleap-optimizely:
@msohailhussain made some small changes to fix the broken tests and make the api more intuitive.

andrewleap-optimizely (approved):
lgtm

msohailhussain (approved):
lgtm

PR #383, opened 2022-05-18
fix: tests that utilize threading from failing with pypy

andrewleap-optimizely:
Summary
-------

-  Fix forced decision test with thread safe call counter
-  Fix polling config manager tests from interfering with their own results
-  Add wrapper class to allow actual testing of the polling thread

Due to a difference in the way pypy implements threading, tests utilizing threading have been intermittently failing the pypy CI. For the polling config manager, depending on when the polling thread started, it would sometimes poison the results, due to the mock library not interacting with the thread correctly.
For the forced decision test, mock's call_count is not thread safe, so the calls were sometimes not counted correctly.

Additionally, some of the polling config manager tests were calling fetch_datafile manually and then confirming that the polling thread was still alive. Since fetch_datafile wasn't being run by the thread, this check was not actually checking what it was purported to check. 

Issues
------

-  Should address these pypy errors:
FAILED tests/test_config_manager.py::PollingConfigManagerTest::test_fetch_datafile__status_exception_raised
      AssertionError: 'New Time' != <MagicMock name='get().headers.get()' id='157266840'>
FAILED tests/test_user_context.py::UserContextTest::test_forced_decision_sync_return_correct_number_of_calls
     AssertionError: 200 != 137

### DIFF
--- a/tests/test_config_manager.py
+++ b/tests/test_config_manager.py
@@ -218,6 +218,36 @@ class StaticConfigManagerTest(base.BaseTest):
         self.assertEqual(1, round(end_time - start_time))
 
 
+class MockPollingConfigManager(config_manager.PollingConfigManager):
+    ''' Wrapper class to allow manual call of fetch_datafile in the polling thread by
+    overriding the _run method.'''
+    def __init__(self, *args, **kwargs):
+        self.run = False
+        self.stop = False
+        super().__init__(*args, **kwargs)
+
+    def _run(self):
+        while self.is_running and not self.stop:
+            if self.run:
+                self.fetch_datafile()
+                self.run = False
+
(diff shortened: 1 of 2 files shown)


Mat001 (approved):
I think it looks ok to me.
I'll re-run pypy tests on gitactions multiple times to see if they are stable.

jaeopt (approved):
LGTM - comments about the test changes will be helpful

msohailhussain:
@andrewleap-optimizely Can you please add expectations of `threading.Thread.start` and `requests.get` that these except these methods are triggered.

msohailhussain (approved):
lgtm. please address comments before merge.

PR #384, opened 2022-06-01
fix: handle datafile provided as bytes

andrewleap-optimizely:
Summary
-------

-  If datafile is provided as bytes, convert to str

The `PollingConfigManager` provides the datafile as bytes.

Test plan
---------
- test_config.test_to_datafile_from_bytes
- test_optimizely_config.test__get_datafile_from_bytes

Issues
------
Currently if the datafile is provided as bytes, including using the default `PollingConfigManager`, it is converted to str incorrectly, resulting in an invalid JSON string being returned from `ProjectConfig.to_datafile` and `OptimizelyConfig.get_datafile`.

Steps to replicate:

```python
client = optimizely.Optimizely(sdk_key=sdk_key)
project_config = client.config_manager.get_config()
datafile = project_config.to_datafile()
json.loads(datafile)
```

### DIFF
--- a/optimizely/project_config.py
+++ b/optimizely/project_config.py
@@ -41,7 +41,7 @@ class ProjectConfig(object):
         """
 
         config = json.loads(datafile)
-        self._datafile = u'{}'.format(datafile)
+        self._datafile = datafile.decode('utf-8') if isinstance(datafile, bytes) else datafile
         self.logger = logger
         self.error_handler = error_handler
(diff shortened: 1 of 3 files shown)


andrewleap-optimizely:
@alexanderchabo we have released this change in [v4.1.0](https://github.com/optimizely/python-sdk/releases/tag/4.1.0)

PR #389, opened 2022-07-05
refactor: type hints private interface

andrewleap-optimizely:
Summary
-------
- Add type hints to the private interface of SDK.
- Enable strict type checking, enforcing type signatures going forward (except on unit tests).

### DIFF
--- a/optimizely/bucketer.py
+++ b/optimizely/bucketer.py
@@ -11,9 +11,18 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
+from typing import Optional, TYPE_CHECKING
 import math
 
+
 from .lib import pymmh3 as mmh3
+
+if TYPE_CHECKING:
+    # prevent circular dependenacy by skipping import at runtime
+    from .project_config import ProjectConfig
+    from .entities import Experiment, Variation
+    from .helpers.types import TrafficAllocation
 
 
 MAX_TRAFFIC_VALUE = 10000
@@ -27,12 +36,12 @@ GROUP_POLICIES = ['random']
 class Bucketer:
     """ Optimizely bucketing algorithm that evenly distributes visitors. """
 
-    def __init__(self):
+    def __init__(self) -> None:
         """ Bucketer init method to set bucketing seed and logger instance. """
 
         self.bucket_seed = HASH_SEED
 
-    def _generate_unsigned_hash_code_32_bit(self, bucketing_id):
+    def _generate_unsigned_hash_code_32_bit(self, bucketing_id: str) -> int:
         """ Helper method to retrieve hash code.
 
         Args:
@@ -45,7 +54,7 @@ class Bucketer:
         # Adjusting MurmurHash code to be unsigned
         return mmh3.hash(bucketing_id, self.bucket_seed) & UNSIGNED_MAX_32_BIT_VALUE
 
-    def _generate_bucket_value(self, bucketing_id):
+    def _generate_bucket_value(self, bucketing_id: str) -> int:
         """ Helper function to generate bucket value in half-closed interval [0, MAX_TRAFFIC_VALUE).
 
         Args:
@@ -58,7 +67,10 @@ class Bucketer:
         ratio = float(self._generate_unsigned_hash_code_32_bit(bucketing_id)) / MAX_HASH_VALUE
         return math.floor(ratio * MAX_TRAFFIC_VALUE)
 
-    def find_bucket(self, project_config, bucketing_id, parent_id, traffic_allocations):
+    def find_bucket(
+        self, project_config: ProjectConfig, bucketing_id: str,
+        parent_id: Optional[str], traffic_allocations: list[TrafficAllocation]
+    ) -> Optional[str]:
         """ Determine entity based on bucket value and traffic allocations.
 
         Args:
@@ -83,7 +95,10 @@ class Bucketer:
 
         return None
 
-    def bucket(self, project_config, experiment, user_id, bucketing_id):
+    def bucket(
+        self, project_config: ProjectConfig,
+        experiment: Experiment, user_id: str, bucketing_id: str
+    ) -> tuple[None | dict[None, None] | Variation, list[str]]:
         """ For a given experiment and bucketing ID determines variation to be shown to user.
 
         Args:
@@ -97,7 +112,7 @@ class Bucketer:
             and array of log messages representing decision making.
      */.
         """
-        decide_reasons = []
+        decide_reasons: list[str] = []
         if not experiment:
             return None, decide_reasons
 
--- a/optimizely/bucketer.py
+++ b/optimizely/bucketer.py
@@ -90,7 +90,7 @@ class Bucketer:
 
         for traffic_allocation in traffic_allocations:
             current_end_of_range = traffic_allocation.get('endOfRange')
-            if bucketing_number < current_end_of_range:
+            if current_end_of_range is not None and bucketing_number < current_end_of_range:
                 return traffic_allocation.get('entityId')
 
         return None
--- a/optimizely/bucketer.py
+++ b/optimizely/bucketer.py
@@ -1,4 +1,4 @@
-# Copyright 2016-2017, 2019-2021 Optimizely
+# Copyright 2016-2017, 2019-2022 Optimizely
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
--- a/optimizely/decision/optimizely_decision.py
+++ b/optimizely/decision/optimizely_decision.py
@@ -45,6 +45,6 @@ class OptimizelyDecision:
             'variables': self.variables,
             'rule_key': self.rule_key,
             'flag_key': self.flag_key,
-            'user_context': self.user_context.as_json(),
+            'user_context': self.user_context.as_json() if self.user_context else None,
             'reasons': self.reasons
         }
--- a/optimizely/decision/optimizely_decision.py
+++ b/optimizely/decision/optimizely_decision.py
@@ -1,4 +1,4 @@
-# Copyright 2021, Optimizely
+# Copyright 2021, 2022, Optimizely
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
--- a/optimizely/decision/optimizely_decision.py
+++ b/optimizely/decision/optimizely_decision.py
@@ -11,10 +11,25 @@
 # See the License for the specific language governing permissions and
 # limitations under the License.
 
+from __future__ import annotations
+from typing import Optional, Any, TYPE_CHECKING
+
+if TYPE_CHECKING:
+    # prevent circular dependenacy by skipping import at runtime
+    from optimizely.optimizely_user_context import OptimizelyUserContext
+
 
 class OptimizelyDecision:
-    def __init__(self, variation_key=None, enabled=None,
-                 variables=None, rule_key=None, flag_key=None, user_context=None, reasons=None):
+    def __init__(
+        self,
+        variation_key: Optional[str] = None,
+        enabled: Optional[bool] = None,
+        variables: Optional[dict[str, Any]] = None,
+        rule_key: Optional[str] = None,
+        flag_key: Optional[str] = None,
+        user_context: Optional[OptimizelyUserContext] = None,
+        reasons: Optional[list[str]] = None
+    ):
         self.variation_key = variation_key
         self.enabled = enabled or False
         self.variables = variables or {}
@@ -23,7 +38,7 @@ class OptimizelyDecision:
         self.user_context = user_context
         self.reasons = reasons or []
 
-    def as_json(self):
+    def as_json(self) -> dict[str, Any]:
         return {
             'variation_key': self.variation_key,
             'enabled': self.enabled,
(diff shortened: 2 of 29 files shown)


optimizely/bucketer.py
```diff
@@ -78,12 +90,15 @@ def find_bucket(self, project_config, bucketing_id, parent_id, traffic_allocatio
...
+            if current_end_of_range is not None and bucketing_number < current_end_of_range:
                 return traffic_allocation.get('entityId')
 
         return None
 
-    def bucket(self, project_config, experiment, user_id, bucketing_id):
+    def bucket(
+        self, project_config: ProjectConfig,
+        experiment: Experiment, user_id: str, bucketing_id: str
+    ) -> tuple[None | dict[None, None] | Variation, list[str]]:
```
msohailhussain, line 101: Is this correct? `dict[None, None]`

andrewleap-optimizely: So `get_variation_from_id_by_experiment_id` and `get_variation_from_key_by_experiment_id` use an empty dict as their null case, which is confusing because a `Variation` is not a dict. I used this type signature to make that more explicit.

I think they should ultimately be changed to return `None` instead.

optimizely/helpers/condition.py
```diff
@@ -486,7 +508,7 @@ def semver_greater_than_evaluator(self, index):
 
         return result > 0
 
-    def semver_less_than_evaluator(self, index):
+    def semver_less_than_evaluator(self, index: int) -> Optional[bool]:
```
msohailhussain, line 511: Why it's optional bool?

optimizely/helpers/condition.py
```diff
@@ -521,7 +543,7 @@ def semver_less_than_evaluator(self, index):
 
         return result < 0
 
-    def semver_less_than_or_equal_evaluator(self, index):
+    def semver_less_than_or_equal_evaluator(self, index: int) -> Optional[bool]:
```
msohailhussain, line 546: same as above.

optimizely/helpers/condition.py
```diff
@@ -606,7 +628,7 @@ def semver_greater_than_or_equal_evaluator(self, index):
         ConditionMatchTypes.SUBSTRING: substring_evaluator
     }
 
-    def split_version(self, version):
+    def split_version(self, version: str) -> Optional[list[str]]:
```
msohailhussain, line 631: why optional?

optimizely/helpers/types.py
```diff
@@ -76,3 +76,25 @@ class RolloutDict(TypedDict):
...
+
+
+class FeatureFlagDict(BaseDict):
+    '''Feature flag dict from parsed datafile json.'''
+    rolloutId: str
+    variables: list[VariableDict]
+    experimentIds: list[str]
+
+
+class GroupDict(TypedDict):
```
msohailhussain, line 88: Is it possible instead of directly inheriting TypedDict, create a BaseEntity ? where we don't have any Id,Key and then we have BaseDict inherity from BaseEntity. 
Use BaseEntity instead of TypedEntity where needed.

optimizely/helpers/validator.py
```diff
@@ -71,7 +80,7 @@ def is_config_manager_valid(config_manager):
     return _has_method(config_manager, 'get_config')
 
 
-def is_event_processor_valid(event_processor):
+def is_event_processor_valid(event_processor: Any) -> bool:
```
msohailhussain, line 83: Please specify type.

msohailhussain: Let's discuss this offline. I know why you mentioned Any.

optimizely/lib/pymmh3.py
```diff
@@ -16,21 +16,21 @@
 
 https://pypi.python.org/pypi/mmh3/2.3.1
 '''
+from __future__ import annotations
```
msohailhussain, line 19: I am leaving this file unreviewed.

optimizely/optimizely.py
```diff
@@ -624,7 +629,7 @@ def get_variation(
 
         return variation_key
 
-    def is_feature_enabled(self, feature_key: str, user_id: str, attributes: Optional[dict[str, Any]] = None) -> bool:
+    def is_feature_enabled(self, feature_key: str, user_id: str, attributes: Optional[UserAttributes] = None) -> bool:
```
msohailhussain, line 632: Can you please confirm, this won't break any code change.

msohailhussain (changes requested):
please address.

optimizely/bucketer.py
msohailhussain: Let's simply it and use tuple with Optional Variation and list of str.

optimizely/decision/optimizely_decision.py
```diff
@@ -11,10 +11,25 @@
...
-    def __init__(self, variation_key=None, enabled=None,
-                 variables=None, rule_key=None, flag_key=None, user_context=None, reasons=None):
+    def __init__(
+        self,
+        variation_key: Optional[str] = None,
+        enabled: Optional[bool] = None,
+        variables: Optional[dict[str, Any]] = None,
+        rule_key: Optional[str] = None,
+        flag_key: Optional[str] = None,
+        user_context: Optional[OptimizelyUserContext] = None,
```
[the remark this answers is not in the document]
andrewleap-optimizely: The only problem with removing the `user_context` default `None`, is that we can't have a non default argument after default arguments. So to remove the `user_context` default, we would need to make `user_context` the first parameter or remove all preceding defaults, which would both be breaking changes.

optimizely/decision_service.py
```diff
@@ -397,7 +435,7 @@ def get_variation_for_rollout(self, project_config, feature, user):
                                                                           bucketing_id)
                 decide_reasons.extend(bucket_reasons)
 
-                if bucketed_variation:
+                if isinstance(bucketed_variation, entities.Variation):
```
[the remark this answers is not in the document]
andrewleap-optimizely: This is a side effect of the dict[None, None] return type. The type checker doesn't know that the dict is always empty and thus that it would never end up in this branch of code. When I change the return type of `get_variation_from_id_by_experiment_id` and `get_variation_from_key_by_experiment_id` to `Optional[Variation]`, this check will no longer be necessary.

optimizely/decision_service.py
```diff
@@ -424,14 +462,19 @@ def get_variation_for_rollout(self, project_config, feature, user):
 
         return Decision(None, None, enums.DecisionSources.ROLLOUT), decide_reasons
 
-    def get_variation_for_feature(self, project_config, feature, user_context, options=None):
+    def get_variation_for_feature(
+        self,
+        project_config: ProjectConfig,
+        feature: entities.FeatureFlag,
+        user_context: OptimizelyUserContext,
+        options: Optional[Sequence[str | OptimizelyDecideOption]] = None
```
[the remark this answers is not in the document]
andrewleap-optimizely: This is my mistake, this can actually only be `str`. `OptimizelyDecideOption` is not a real enum, so it can't be used as the type (without being converted).

optimizely/entities.py
```diff
@@ -41,18 +57,18 @@ def __init__(self, id, key, experimentIds, **kwargs):
...
-        layerId,
-        audienceConditions=None,
-        groupId=None,
-        groupPolicy=None,
-        **kwargs
+        id: str,
+        key: str,
+        status: str,
+        audienceIds: list[str],
+        variations: list[VariationDict],
```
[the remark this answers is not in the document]
andrewleap-optimizely: This class is mirroring the names from the datafile, so changing this field name is non trivial. Instead of being able to auto unpack the datafile object, we would have to manually assign in a couple different places. For example:
```diff
 for experiment_dict in layer.experiments:
-     self.experiment_id_map[experiment_dict['id']] = entities.Experiment(**experiment_dict)
+     self.experiment_id_map[experiment_dict['id']] = entities.Experiment(
+                    id=experiment_dict['id'],
+                    key=experiment_dict['key'],
+                    status=experiment_dict['status'],
+                    audienceIds=experiment_dict['audienceIds'],
+                    variationsDict=experiment_dict['variations'],
+                    forcedVariations=experiment_dict['forcedVariations'],
+                    trafficAllocation=experiment_dict['trafficAllocation'],
+                    layerId=experiment_dict['layerId'],
+                    audienceConditions=experiment_dict.get('audienceConditions'),
+                    groupId=cast(Optional[str], experiment_dict.get('groupId')),
+                    groupPolicy=cast(Optional[str], experiment_dict.get('groupPolicy'))
+                )
```

Might not be worth it?

optimizely/event/event_factory.py
```diff
@@ -28,12 +43,16 @@ class EventFactory:
   """
 
     EVENT_ENDPOINT = 'https://logx.optimizely.com/v1/events'
-    HTTP_VERB = 'POST'
+    HTTP_VERB: Final = 'POST'
```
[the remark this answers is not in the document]
andrewleap-optimizely: makes sense

optimizely/event_builder.py
```diff
@@ -35,7 +55,7 @@ class EventBuilder:
   impressions and conversions using the new V3 event API (batch). """
 
     EVENTS_URL = 'https://logx.optimizely.com/v1/events'
-    HTTP_VERB = 'POST'
+    HTTP_VERB: Final = 'POST'
```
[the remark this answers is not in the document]
andrewleap-optimizely: agreed

optimizely/helpers/condition.py
andrewleap-optimizely: Doc string says:
https://github.com/optimizely/python-sdk/blob/9b4ec4f109cf4eac33b64e7fc425278e73f7f420/optimizely/helpers/condition.py#L517-L522

optimizely/helpers/condition.py
andrewleap-optimizely: Looks like a similar reason:
https://github.com/optimizely/python-sdk/blob/9b4ec4f109cf4eac33b64e7fc425278e73f7f420/optimizely/helpers/condition.py#L552-L557

optimizely/helpers/condition.py
andrewleap-optimizely: Doc string:
https://github.com/optimizely/python-sdk/blob/9b4ec4f109cf4eac33b64e7fc425278e73f7f420/optimizely/helpers/condition.py#L637-L641

optimizely/helpers/types.py
andrewleap-optimizely: Yea that should work

optimizely/optimizely.py
andrewleap-optimizely: This is essentially a no-op wrapper, that gives it a name for the type checker to track. But at runtime the behavior is unchanged.

https://docs.python.org/3/library/typing.html#typing.NewType

optimizely/helpers/audience.py
```diff
@@ -11,18 +11,28 @@
...
+if TYPE_CHECKING:
+    # prevent circular dependenacy by skipping import at runtime
+    from optimizely.project_config import ProjectConfig
+    from optimizely.logger import Logger
+
+
+def does_user_meet_audience_conditions(
+    config: ProjectConfig,
+    audience_conditions: Optional[Sequence[str | list[str]]],
+    audience_logs: Any,
```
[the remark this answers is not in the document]
andrewleap-optimizely: Except `audience_conditions` can also be a list of lists:
```python
['and', ['or', '1234'], ['or', '5678']]
```

The type should be something like:
```python
Condition = Optional[str | list[Condition]]
```
but unfortunately recursive types aren't yet supported. I think a list of lists is as close to an accurate representation as we can get.

optimizely/event_builder.py
```diff
@@ -125,7 +149,7 @@ def _get_common_params(self, project_config, user_id, attributes):
     Returns:
      Dict consisting of parameters common to both impression and conversion events.
     """
-        common_params = {
+        common_params: dict[str, Any] = {
```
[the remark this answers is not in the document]
andrewleap-optimizely: Here's an example value from the test suite:
```python
{
    "project_id": "111001",
    "account_id": "12001",
    "visitors": [
        {
            "visitor_id": "test_user",
            "snapshots": [],
            "attributes": []
        }
    ],
    "client_name": "python-sdk",
    "enrich_decisions": True,
    "client_version": "4.0.0",
    "anonymize_ip": False,
    "revision": "42"
}
```
To really make this one work, we would need to use a bunch of TypedDicts; Although I tried it and based on the way these dicts are constructed, it becomes quite obnoxious to setup, and would require us adding a dependency for python 3.8+.

optimizely/event_builder.py
```diff
@@ -159,7 +185,7 @@ def _get_required_params_for_impression(self, experiment, variation_id):
     Returns:
       Dict consisting of decisions and events info for impression event.
     """
-        snapshot = {}
+        snapshot: dict[str, list[dict[str, str | int]]] = {}
```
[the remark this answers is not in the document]
andrewleap-optimizely: The Snapshot class cannot be swapped in here without a significant refactor, unless you're talking about a different class?

msohailhussain (approved):
lgtm

optimizely/helpers/audience.py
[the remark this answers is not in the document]
andrewleap-optimizely: My mistake, I updated the type for `audience_logs`

optimizely/helpers/validator.py
andrewleap-optimizely: added types to validator functions with known expected types

Mat001 (approved):
Looks good to me. Hope I didn't miss anything, I went through most of it before.

PR #398, opened 2022-08-08
feat: add odp rest api manager

Mat001:
Summary
-------

-  Added rest api code for odp. Sending events to ODP platform.
- added class OdpEvent
- added a helper function for creating and mocking a fake server response
- added unit tests
- fixed typo in enums
- minor fixes to odp graphql manager tests
- See acceptance criteria in the ticket below.

Test plan
---------
- unit tests
- FSC tests

Issues
------

-  [OASIS-8447 Python - add ODP RestApi interface](https://optimizely.atlassian.net/browse/OASIS-8447)


optimizely/odp/odp_event.py
```diff
@@ -0,0 +1,27 @@
...
+
+from typing import Any, Dict
+
+
+class OdpEvent:
+    """ Representation of an odp event which can be sent to the Optimizely odp platform. """
+
+    def __init__(self, type: str, action: str,
+                 identifiers: Dict[str, str], data: Dict[str, Any]) -> None:
+        self.type = type,
```
Mat001, line 24: @andrewleap-optimizely I'm shadowing the keyword "type". But Swift uses it. What u think? It's probably dafe as it's encapsulated in the class?

optimizely/odp/zaius_rest_api_manager.py
```diff
@@ -0,0 +1,86 @@
...
+            can_retry = False
+        except RequestException as err:
+            if 400 <= err.response.status_code < 500:
+                self.logger.error(Errors.ODP_EVENT_FAILED.format(err))
+                can_retry = False
+            else:
+                self.logger.error(Errors.ODP_EVENT_FAILED.format(err))
+                # we do retry, can_retry = True
+        finally:
+            return can_retry
```
Mat001, line 86: I wonder if default return should be to NOT retry?

tests/helpers_for_tests.py
```diff
@@ -0,0 +1,25 @@
...
+# distributed under the License is distributed on an "AS IS" BASIS,
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+
+from requests import Response
+from typing import Optional
+
+
+def fake_server_response(status_code: int = None, content: bytes = None, url: str = None) -> Optional[Response]:
```
Mat001, line 18: @andrewleap-optimizely What do you think here, I got content as type bytes, correct?

Mat001: Also @andrewleap-optimizely I'm not happy with the file name "helpers_for_tests". There is helpers in optimizely folder, so wanted to be consistent, but can't start the file with prefix test_ (as in test_helpers). Could use testutils.py, utils.py... suggestions?

optimizely/odp/zaius_rest_api_manager.py
Mat001: @andrewleap-optimizely I'm returning True or False. I forgot if we said to return one of the boolean (True/falsE) for retrying and then None for not. Correct me if I'm wrong. Lmk if True/False works for your OdpEventManager

tests/helpers_for_tests.py
Mat001: > @andrewleap-optimizely What do you think here, I got content as type bytes, correct?

Changed bytes to str

optimizely/odp/zaius_rest_api_manager.py
andrewleap-optimizely: Yea it should be a non optional bool, otherwise it looks sound.

However you should remove that finally and just have a bare return. Finally with a return will discard any uncaught exceptions

tests/helpers_for_tests.py
andrewleap-optimizely: What about adding it as a method to `BaseTest` in base.py? Then it can be called easily with `self.fake_server_response` in any test

optimizely/odp/zaius_rest_api_manager.py
```diff
@@ -0,0 +1,86 @@
...
+                                     data=json.dumps(events),
+                                     timeout=OdpRestApiConfig.REQUEST_TIMEOUT)
+
+            response.raise_for_status()
+            can_retry = False
+
+        except (ConnectionError, Timeout):
+            self.logger.error(Errors.ODP_EVENT_FAILED.format('network error'))
+            # we do retry, can_retry = True
+        except JSONDecodeError:
```
andrewleap-optimizely, line 74: Since there's no call to `response.json()` we don't need this except

optimizely/odp/zaius_rest_api_manager.py
```diff
@@ -0,0 +1,86 @@
...
+            self.logger.error(Errors.ODP_EVENT_FAILED.format('JSON decode error'))
+            can_retry = False
+        except InvalidURL:
+            self.logger.error(Errors.ODP_EVENT_FAILED.format('invalid URL'))
+            can_retry = False
+        except RequestException as err:
+            if 400 <= err.response.status_code < 500:
+                self.logger.error(Errors.ODP_EVENT_FAILED.format(err))
```
andrewleap-optimizely, line 82: You should log the body of the response here and fall back to the exception if there is no body.

andrewleap-optimizely (changes requested):
Looks good so far! I like that you reused the mock response, that's a good idea. Just a couple changes

optimizely/odp/zaius_rest_api_manager.py
```diff
@@ -0,0 +1,88 @@
...
+            retry is True - if network or server error (5xx), otherwise False
+        """
+        can_retry: bool = True
+        url = f'{api_host}/v3/events'
+        request_headers = {'content-type': 'application/json', 'x-api-key': api_key}
+
+        try:
+            response = requests.post(url=url,
+                                     headers=request_headers,
+                                     data=json.dumps(events),
```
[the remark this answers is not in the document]
Mat001: Great catch, thx.

jaeopt (changes requested):
Looks good. A couple of small changes suggested.

optimizely/odp/zaius_rest_api_manager.py
Mat001: @andrewleap-optimizely Do you mean the response body from the request (not from the exc)?
What were you thinking, that there can be a body that is not an exception?

I tested with a few 4xx errors and logged response body from the try block and none of then had any content.
Only "err" - the exception itself seems to contain the error message.

Mat001: I wasn't mocking the response content. Worked when I added it.

Mat001:
@jaeopt @andrewleap-optimizely I hope exception for invalid URL is handled ok. Request.Exception's class InvalidURL is not comprehensive at all. For example it doesn't catch https://api.zaius..com or XXhttps://api.zaius.com. For cases like these two I had to add other exceptions. If better idea how to guard against an invalid url lmk

I'm not super clear how swift does it, it seems it checks for exactly that url string?
`guard let url = URL(string: "\(apiHost)/v3/events") else {`
`            let canRetry = false`
`            completionHandler(.odpEventFailed("Invalid url", canRetry))`
`            return`
`        }`

andrewleap-optimizely:
> I'm not super clear how swift does it, it seems it checks for exactly that url string? `guard let url = URL(string: "\(apiHost)/v3/events") else {` ` let canRetry = false` ` completionHandler(.odpEventFailed("Invalid url", canRetry))` ` return` ` }`

We can accomplish something similar using `urllib.parse.urlparse`

jaeopt:
> @jaeopt @andrewleap-optimizely I hope exception for invalid URL is handled ok. Request.Exception's class InvalidURL is not comprehensive at all. For example it doesn't catch https://api.zaius..com or XXhttps://api.zaius.com. For cases like these two I had to add other exceptions. If better idea how to guard against an invalid url lmk
> 
> I'm not super clear how swift does it, it seems it checks for exactly that url string? `guard let url = URL(string: "\(apiHost)/v3/events") else {` ` let canRetry = false` ` completionHandler(.odpEventFailed("Invalid url", canRetry))` ` return` ` }`

@Mat001 I understand that RequestException will catch these all.
I don't think we need this fine level of error handling, since this is not expected to happen (apiHost we control).
Can be cleaned up if needed. Swift checks url only for lang level requirement for optional filtering.

Mat001:
@jaeopt @andrewleap-optimizely  ok. I refactored. I got rid of the specific URL parsing exceptions and made it such that generic RequestEception catches invalid URL errors (may not include urlparse errors), but it guard against quite a few. And since we control the URL I wouldn't worry about this one, if that's ok

optimizely/odp/zaius_rest_api_manager.py
```diff
@@ -55,34 +55,40 @@ def sendOdpEvents(self, api_key: str, api_host: str, events: List[OdpEvent]) ->
...
-            # we do retry, can_retry = True
-        except JSONDecodeError:
-            self.logger.error(Errors.ODP_EVENT_FAILED.format('JSON decode error'))
-            can_retry = False
-        except InvalidURL:
-            self.logger.error(Errors.ODP_EVENT_FAILED.format('invalid URL'))
-            can_retry = False
+            # retry on network errors
+            should_retry = True
         except RequestException as err:
```
jaeopt, line 80: I think so, but can you confirm that this catches all other exceptions?

Mat001: @jaeopt yes, confirm. We use the same exception handling in event_dispatcher and fetch_datafile:
https://github.com/optimizely/python-sdk/blob/master/optimizely/event_dispatcher.py#L56 
https://github.com/optimizely/python-sdk/blob/master/optimizely/config_manager.py#L367

It'll catch all these: https://requests.readthedocs.io/en/latest/_modules/requests/exceptions/

andrewleap-optimizely (approved):
lgtm

PR #401, opened 2022-08-19
feat: add odp config

andrewleap-optimizely:
Summary
-------
Config for managing odp integration

Test plan
---------
- added test_odp_config.py

Ticket
---------
[OASIS-8405](https://optimizely.atlassian.net/browse/OASIS-8405)
[OASIS-8442](https://optimizely.atlassian.net/browse/OASIS-8442)

### DIFF
--- a/optimizely/odp/odp_config.py
+++ b/optimizely/odp/odp_config.py
@@ -73,7 +73,7 @@ class OdpConfig:
         with self.lock:
             return self.api_key
 
-    def set_api_key(self, api_key: Optional[str]):
+    def set_api_key(self, api_key: Optional[str]) -> None:
         with self.lock:
             self.api_key = api_key
 
@@ -81,7 +81,7 @@ class OdpConfig:
         with self.lock:
             return self.segments_to_check.copy()
 
-    def set_segments_to_check(self, segments_to_check: list[str]):
+    def set_segments_to_check(self, segments_to_check: list[str]) -> None:
         """
         Replace qualified segments with provided list of segments
         Args:
--- a/optimizely/odp/odp_config.py
+++ b/optimizely/odp/odp_config.py
@@ -24,7 +24,7 @@ class OdpConfig:
     Args:
         api_host: The host URL for the ODP audience segments API (optional).
         api_key: The public API key for the ODP account from which the audience segments will be fetched (optional).
-        segments_to_check: An array of all ODP segments used in the current datafile
+        segments_to_check: A list of all ODP segments used in the current datafile
                             (associated with api_host/api_key).
     """
     def __init__(
@@ -45,7 +45,7 @@ class OdpConfig:
         Args:
             api_host: The host URL for the ODP audience segments API (optional).
             api_key: The public API key for the ODP account from which the audience segments will be fetched (optional).
-            segments_to_check: An array of all ODP segments used in the current datafile
+            segments_to_check: A list of all ODP segments used in the current datafile
                                (associated with api_host/api_key).
 
         Returns:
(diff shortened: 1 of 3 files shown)


optimizely/odp/odp_config.py
```diff
@@ -0,0 +1,96 @@
...
+        self.lock = Lock()
+
+    def update(self, api_key: Optional[str], api_host: Optional[str], segments_to_check: list[str]) -> bool:
+        """
+        Override the ODP configuration.
+
+        Args:
+            api_host: The host URL for the ODP audience segments API (optional).
+            api_key: The public API key for the ODP account from which the audience segments will be fetched (optional).
+            segments_to_check: An array of all ODP segments used in the current datafile
```
Mat001, line 48: Maybe change "array" to "list"?

optimizely/odp/odp_config.py
```diff
@@ -0,0 +1,96 @@
...
+                               (associated with api_host/api_key).
+
+        Returns:
+            True if the provided values were different than the existing values.
+        """
+        updated = False
+        with self.lock:
+            if self.api_key != api_key or self.api_host != api_host or self.segments_to_check != segments_to_check:
```
Mat001, line 56: I prefer to use "not" as in more pythonic....what u think?
Visually ! looks better?

optimizely/odp/odp_config.py
```diff
@@ -0,0 +1,96 @@
...
+        with self.lock:
+            self.api_key = api_key
+
+    def get_segments_to_check(self) -> Optional[list[str]]:
+        with self.lock:
+            return self.segments_to_check.copy()
+
+    def set_segments_to_check(self, segments_to_check: list[str]) -> None:
+        """
+        Replace qualified segments with provided list of segments
```
Mat001, line 86: Minor...I think we could use proper sentence with period etc at the end. The rest os SDK does that. Maybe for consistency? Then we can discuss if generally we care about periods or not? (goes for all comments in the file)

optimizely/odp/odp_config.py
```diff
@@ -0,0 +1,96 @@
...
+        Args:
+            api_host: The host URL for the ODP audience segments API (optional).
+            api_key: The public API key for the ODP account from which the audience segments will be fetched (optional).
+            segments_to_check: An array of all ODP segments used in the current datafile
+                               (associated with api_host/api_key).
+
+        Returns:
+            True if the provided values were different than the existing values.
+        """
+        updated = False
```
Mat001, line 54: Do we need "updated" var here, why not returning True/false directly?
I might agree though that using "update" is more readable.

Mat001 (approved):
Few minor comments.

optimizely/odp/odp_config.py
andrewleap-optimizely: Yea I think naming the boolean makes the intention clearer. Also this goes back to a [conversation](https://github.com/optimizely/ruby-sdk/pull/310#discussion_r948297685) I had with Zeeshan about returning from a synchronized block. I'm almost 100% sure it's not a problem in python, but might be better to not distract future devs with that worry 😄.

optimizely/odp/odp_config.py
andrewleap-optimizely: I think it reads more like english with `self.api_key != api_key` vs `not self.api_key == api_key`. I think `not` works better with identity comparisons ie  `self.api_key is not api_key` or flipping a bool `not updated`.

Also adding three nots would push us over the line limit 😃

Mat001: I agree. Python's "not" is not positioned intuitively in equality evaluation

optimizely/odp/odp_config.py
```diff
@@ -0,0 +1,96 @@
...
+            return self.segments_to_check.copy()
+
+    def set_segments_to_check(self, segments_to_check: list[str]) -> None:
+        """
+        Replace qualified segments with provided list of segments.
+        Args:
+          segments_to_check: A list of segment names.
+        """
+        with self.lock:
+            self.segments_to_check = segments_to_check.copy()
```
jaeopt, line 91: I do not see any case to set these props separately other than updating altogether with update(). Any use case in mind?

jaeopt (approved):
LGTM with a possible clean up

optimizely/odp/odp_config.py
andrewleap-optimizely: Nope! Removed.

PR #403, opened 2022-08-23
feat: add odp event manager

andrewleap-optimizely:
Summary
-------
OdpEventManager provides internal services for sending ODP events to the ODP server. 
- adds default values to identifiers and data fields for all events.
- stores requested events if they cannot be sent to the server.
- stored events are dispatched when resources are available.

Test plan
---------
- added test_odp_event_manager.py

Ticket
------
[OASIS-8442](https://optimizely.atlassian.net/browse/OASIS-8442)

### DIFF
--- a/optimizely/odp/odp_event_manager.py
+++ b/optimizely/odp/odp_event_manager.py
@@ -12,11 +12,11 @@
 # limitations under the License.
 
 from __future__ import annotations
+from enum import Enum
 from threading import Thread
 from typing import Any, Optional
-import queue
-from queue import Queue
-from sys import version_info
+import time
+from queue import Empty, Queue, Full
 
 from optimizely import logger as _logging
 from .odp_event import OdpEvent
@@ -25,15 +25,10 @@ from .zaius_rest_api_manager import ZaiusRestApiManager
 from optimizely.helpers.enums import OdpEventManagerConfig, Errors
 
 
-if version_info < (3, 8):
-    from typing_extensions import Final
-else:
-    from typing import Final  # type: ignore
-
-
-class Signal:
-    '''Used to create unique objects for sending signals to event queue.'''
-    pass
+class Signal(Enum):
+    '''Enum for sending signals for the event queue.'''
+    SHUTDOWN = 1
+    FLUSH = 2
 
 
 class OdpEventManager:
@@ -44,15 +39,11 @@ class OdpEventManager:
     the queue and buffers them before the events are sent to ODP.
     """
 
-    _SHUTDOWN_SIGNAL: Final = Signal()
-    _FLUSH_SIGNAL: Final = Signal()
-
     def __init__(
         self,
         odp_config: OdpConfig,
         logger: Optional[_logging.Logger] = None,
         api_manager: Optional[ZaiusRestApiManager] = None
-
     ):
         """ OdpEventManager init method to configure event batching.
 
@@ -66,59 +57,87 @@ class OdpEventManager:
         self.odp_config = odp_config
         self.event_queue: Queue[OdpEvent | Signal] = Queue(OdpEventManagerConfig.DEFAULT_QUEUE_CAPACITY)
         self.batch_size = OdpEventManagerConfig.DEFAULT_BATCH_SIZE
+        self.flush_interval = OdpEventManagerConfig.DEFAULT_FLUSH_INTERVAL
+        self._set_flush_deadline()
         self._current_batch: list[OdpEvent] = []
-        self.executor = Thread(target=self._run, daemon=True)
+        """_current_batch should only be modified by the processing thread, as it is not thread safe"""
+        self.thread = Thread(target=self._run, daemon=True)
+        self.thread_exception = False
+        """thread_exception will be True if the processing thread did not exit cleanly"""
 
     @property
     def is_running(self) -> bool:
         """ Property to check if consumer thread is alive or not. """
-        return self.executor.is_alive()
+        return self.thread.is_alive()
 
     def start(self) -> None:
         """ Starts the batch processing thread to batch events. """
         if self.is_running:
-            self.logger.warning('ODP event processor already started.')
+            self.logger.warning('ODP event queue already started.')
             return
 
-        self.executor.start()
+        self.thread.start()
 
     def _run(self) -> None:
         """ Triggered as part of the thread which batches odp events or flushes event_queue and blocks on get
         for flush interval if queue is empty.
         """
         try:
-            while True:
-                item = self.event_queue.get()
+            item = None
+            self.odp_config.odp_ready.wait()
+            self.logger.debug('ODP ready. Starting event processing.')
 
-                if item == self._SHUTDOWN_SIGNAL:
+            while True:
+                timeout = self._get_time_till_flush()
+
+                try:
+                    item = self.event_queue.get(True, timeout)
+                except Empty:
+                    item = None
+
+                if item == Signal.SHUTDOWN:
                     self.logger.debug('Received ODP event shutdown signal.')
-                    self.event_queue.task_done()
                     break
 
-                if item is self._FLUSH_SIGNAL:
+                elif item == Signal.FLUSH:
                     self.logger.debug('Received ODP event flush signal.')
                     self._flush_batch()
                     self.event_queue.task_done()
                     continue
 
-                if isinstance(item, OdpEvent):
+                elif isinstance(item, OdpEvent):
                     self._add_to_batch(item)
                     self.event_queue.task_done()
 
+                elif len(self._current_batch) > 0:
+                    self.logger.debug('Flushing on interval.')
+                    self._flush_batch()
+
+                else:
+                    self._set_flush_deadline()
+
         except Exception as exception:
+            self.thread_exception = True
             self.logger.error(f'Uncaught exception processing ODP events. Error: {exception}')
 
         finally:
             self.logger.info('Exiting ODP event processing loop. Attempting to flush pending events.')
             self._flush_batch()
+            if item == Signal.SHUTDOWN:
+                self.event_queue.task_done()
 
     def flush(self) -> None:
         """ Adds flush signal to event_queue. """
-
-        self.event_queue.put(self._FLUSH_SIGNAL)
+        try:
+            self.event_queue.put_nowait(Signal.FLUSH)
+        except Full:
+            self.logger.error("Error flushing ODP event queue")
 
     def _flush_batch(self) -> None:
-        """ Flushes current batch by dispatching event. """
+        """ Flushes current batch by dispatching event.
+        Should only be called by the executor thread."""
+        self._set_flush_deadline()
+
         batch_len = len(self._current_batch)
         if batch_len == 0:
             self.logger.debug('Nothing to flush.')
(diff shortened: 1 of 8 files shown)


optimizely/odp/odp_event_manager.py
```diff
@@ -0,0 +1,186 @@
...
+                    self.logger.debug('Received ODP event flush signal.')
+                    self._flush_batch()
+                    self.event_queue.task_done()
+                    continue
+
+                if isinstance(item, OdpEvent):
+                    self._add_to_batch(item)
+                    self.event_queue.task_done()
+
+        except Exception as exception:
```
Mat001, line 108: I know existing BatchEventProcessor uses this catch-all. I wonder we could use a less generic class of exceptions, however I can see it wouldn't be straightforward to isolate them (issues with signaling, flushing, adding to batch,...). Maybe capturing all is desirable here nevertheless.

tests/test_odp_event_manager.py
```diff
@@ -0,0 +1,273 @@
...
+            event_manager.send_event(**self.events[1])
+            event_manager.event_queue.join()
+
+        self.assertEqual(mock_send.call_count, 2)
+        for call in mock_send.call_args_list:
+            self.assertEqual(call[0], (self.api_key, self.api_host, self.processed_events))
+
+        self.assertEqual(len(event_manager._current_batch), 0)
+        mock_logger.error.assert_not_called()
+        mock_logger.debug.assert_any_call('Flushing ODP events on batch size.')
```
Mat001, line 136: Could you assert here that we get logs twice, one for each flush?

tests/test_odp_event_manager.py
```diff
@@ -0,0 +1,273 @@
...
+            event_manager.send_event(**self.events[1])
+            event_manager.flush()
+            event_manager.event_queue.join()
+
+        self.assertEqual(mock_send.call_count, 2)
+        for call in mock_send.call_args_list:
+            self.assertEqual(call[0], (self.api_key, self.api_host, self.processed_events))
+        mock_logger.error.assert_not_called()
+        self.assertEqual(len(event_manager._current_batch), 0)
+        mock_logger.debug.assert_any_call('Received ODP event flush signal.')
```
Mat001, line 179: add mock call we received 2 logs, one for each flush?

tests/test_odp_event_manager.py
```diff
@@ -0,0 +1,273 @@
...
+            event_manager.send_event(**self.events[1])
+            event_manager.event_queue.join()
+
+        mock_send.assert_called_once_with(self.api_key, self.api_host, self.processed_events)
+        self.assertEqual(len(event_manager._current_batch), 0)
+        mock_logger.error.assert_not_called()
+        mock_logger.debug.assert_any_call('Flushing ODP events on batch size.')
+        event_manager.stop()
+
+    def test_odp_event_manager_multiple_batches(self):
```
Mat001, line 115: I like this test on multiple batches, and the one after, for multiple flushes. There are always going to be many batches flushed in real scenarios.

tests/test_odp_event_manager.py
```diff
@@ -0,0 +1,273 @@
...
+        {
+            "type": "t1",
+            "action": "a1",
+            "identifiers": {"id-key-1": "id-value-1"},
+            "data": {
+                "idempotence_id": test_uuid,
+                "data_source_type": "sdk",
+                "data_source": "python-sdk",
+                "data_source_version": __version__,
+                "key-1": "value1"
```
jaeopt, line 58: Can we have a test case that custom data overwrites the default data like {"data_source": "my-app"}

jaeopt (changes requested):
Looks good. A few changes suggested. 
Also let's discuss how we should handle events with odpConfig lifecycle.

optimizely/odp/odp_event_manager.py
```diff
@@ -0,0 +1,186 @@
...
+            self.logger.debug('Error dispatching ODP events, scheduled to retry.')
+            return
+
+        self._current_batch = []
+
+    def _add_to_batch(self, odp_event: OdpEvent) -> None:
+        """ Method to append received odp event to current batch."""
+
+        self._current_batch.append(odp_event)
+        if len(self._current_batch) >= self.batch_size:
```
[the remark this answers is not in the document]
andrewleap-optimizely: Yea until shutdown... should we instead implement a flush interval like the `BatchEventProcessor` or flush every event, using batching essentially only for retries?

optimizely/odp/odp_event_manager.py
```diff
@@ -0,0 +1,186 @@
...
+        if not self.odp_config.odp_integrated():
+            self.logger.debug('ODP event processing has been disabled.')
+            return
+
+        if not self.is_running:
+            self.logger.warning('ODP event processor is shutdown, not accepting events.')
+            return
+
+        try:
+            self.event_queue.put_nowait(event)
```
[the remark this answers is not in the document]
andrewleap-optimizely: event_queue is thread safe. current_batch is not, but it is only accessed by the batch processing thread.

optimizely/odp/odp_event_manager.py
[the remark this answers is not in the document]
jaeopt: @andrewleap-optimizely yes, we can consider adding a timeout here too. For now, we do not need to make it configurable. The ODP events may be a bit more time-sensitive so timeout should be equal or less than the existing batch timeout.

optimizely/odp/odp_event_manager.py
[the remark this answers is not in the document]
jaeopt: @andrewleap-optimizely a comment for this clarification may be helpful.

optimizely/odp/odp_event_manager.py
```diff
@@ -0,0 +1,187 @@
...
+            self.logger.error('Error stopping ODP event queue.')
+
+    def send_event(self, type: str, action: str, identifiers: dict[str, str], data: dict[str, Any]) -> None:
+        event = OdpEvent(type, action, identifiers, data)
+        self.dispatch(event)
+
+    def dispatch(self, event: OdpEvent) -> None:
+        if not self.odp_config.odp_integrated():
+            self.logger.debug('ODP event processing has been disabled.')
+            return
```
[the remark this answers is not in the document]
andrewleap-optimizely: I was thinking we would just block on datafile ready when calling `OptimizelyUserContext.fetch_qualified_segments`, so that becomes a non issue. Especially since on server, Optimizely client initialization should(?) be an infrequent occurrence.

If you think that's not a good solution, I can set it up similarly to how you have it implemented in swift?

optimizely/odp/odp_event_manager.py
```diff
@@ -0,0 +1,187 @@
...
+        if batch_len == 0:
+            self.logger.debug('Nothing to flush.')
+            return
+
+        api_key = self.odp_config.get_api_key()
+        api_host = self.odp_config.get_api_host()
+
+        if not api_key or not api_host:
+            self.logger.debug('ODP event processing has been disabled.')
+            return
```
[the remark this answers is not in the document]
andrewleap-optimizely: I suppose discard would make more sense here, assuming that the datafile has already been initialized at this point.

optimizely/odp/odp_event_manager.py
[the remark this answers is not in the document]
andrewleap-optimizely: I agree!

optimizely/odp/odp_event_manager.py
andrewleap-optimizely: Since exceptions in child threads don't propagate to the main thread, we should keep this exception. I also added a flag so that the main thread can throw errors on subsequent attempted send events.

tests/base.py
```diff
@@ -14,12 +14,25 @@
...
 from typing import Optional
+from copy import deepcopy
+from unittest import mock
 
 from requests import Response
 
 from optimizely import optimizely
 
 
+class CopyingMock(mock.MagicMock):
```
Mat001, line 25: Is CopyingMock meat as a verb or as a noun?
I wonder if we could have it as a noun, like all classes, just so we have the same flavor to class names?
For example "CopiedMock" or similar?
Thoughts?

Mat001 (approved):
Looks good to me. Pretty complex PR.
Made a comment about the name CopyingMock class.

optimizely/odp/odp_event_manager.py
```diff
@@ -0,0 +1,233 @@
...
+
+    def send_event(self, type: str, action: str, identifiers: dict[str, str], data: dict[str, Any]) -> None:
+        """Create OdpEvent and add it to the event queue."""
+        odp_state = self.odp_config.odp_state()
+        if odp_state == OdpConfigState.UNDETERMINED:
+            self.logger.debug('ODP events cannot be sent before the datafile has loaded.')
+            return
+
+        if odp_state == OdpConfigState.NOT_INTEGRATED:
+            self.logger.debug('ODP event queue has been disabled.')
```
jaeopt, line 208:
```suggestion
            self.logger.debug('ODP is not integrated.')
```

optimizely/odp/odp_event_manager.py
```diff
@@ -0,0 +1,233 @@
...
+        batch_len = len(self._current_batch)
+        if batch_len == 0:
+            self.logger.debug('Nothing to flush.')
+            return
+
+        api_key = self.odp_config.get_api_key()
+        api_host = self.odp_config.get_api_host()
+
+        if not api_key or not api_host:
+            self.logger.debug('ODP event queue has been disabled.')
```
jaeopt, line 147:
```suggestion
            self.logger.debug('ODP is not integrated.')
```
Can we add this error to enums like other standard error messages?

tests/test_odp_event_manager.py
```diff
@@ -0,0 +1,434 @@
...
+        mock_logger = mock.Mock()
+
+        with mock.patch('optimizely.helpers.enums.OdpEventManagerConfig.DEFAULT_QUEUE_CAPACITY', 1):
+            event_manager = OdpEventManager(self.odp_config, mock_logger)
+
+        with mock.patch('optimizely.odp.odp_event_manager.OdpEventManager.is_running', True):
+            event_manager.send_event(**self.events[0])
+            event_manager.send_event(**self.events[1])
+
+        mock_logger.warning.assert_called_once_with('ODP event send failed (Queue is full).')
```
jaeopt, line 301: Is this "warning" (not "error") a good level? Can you check if this is consistent with logx event overflow?

jaeopt (changes requested):
Looks great! A few changes suggested.

tests/test_odp_event_manager.py
andrewleap-optimizely: This is consistent with logx event overflow. However since the event is dropped, error level doesn't seem out of place, in my opinion.

optimizely/odp/odp_event_manager.py
```diff
@@ -66,59 +59,86 @@ def __init__(
...
         try:
+            self.odp_config.odp_ready.wait()
+            self.logger.debug('ODP ready. Starting event processing.')
+
             while True:
-                item = self.event_queue.get()
+                timeout = self._get_time_till_flush()
+
+                try:
+                    item = self.event_queue.get(True, timeout)
```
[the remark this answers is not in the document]
andrewleap-optimizely: Changed the logic so countdown starts on first event added to queue.

tests/test_odp_event_manager.py
jaeopt: @andrewleap-optimizely if it's consistent with logx overflow, let's keep it consistent. error or warning is debatable since overflow may be a result of other errors.

jaeopt (changes requested):
All changes look good. 
I found one more case to discuss to clarify.

jaeopt (approved):
LGTM

PR #410, opened 2022-11-08
feat: add stop method to polling config manager

andrewleap-optimizely:
Summary
-------

- Add stop method to PollingConfigManager
- Stop PollingConfigManager from Optimizely.close()
- Fix Optimizely docstring to say BatchEventProcessor is the default
- Lock mypy to version 0.982 due to regression

Test plan
---------
fixed tests

Ticket
------

[FSSDK-8499](https://jira.sso.episerver.net/browse/FSSDK-8499)

### DIFF
--- a/optimizely/optimizely.py
+++ b/optimizely/optimizely.py
@@ -88,10 +88,9 @@ class Optimizely:
                                config_manager.BaseConfigManager implementation which can be using the
                                same NotificationCenter instance.
           event_processor: Optional component which processes the given event(s).
-                           By default optimizely.event.event_processor.ForwardingEventProcessor is used
-                           which simply forwards events to the event dispatcher.
-                           To enable event batching configure and use
-                           optimizely.event.event_processor.BatchEventProcessor.
+                           By default optimizely.event.event_processor.BatchEventProcessor is used
+                           which batches events. To simply forward events to the event dispatcher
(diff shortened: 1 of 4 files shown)


optimizely/optimizely.py
```diff
@@ -1381,3 +1380,5 @@ def close(self) -> None:
         if callable(getattr(self.event_processor, 'stop', None)):
             self.event_processor.stop()  # type: ignore[attr-defined]
         self.odp_manager.close()
+        if callable(getattr(self.config_manager, 'stop', None)):
+            self.config_manager.stop()  # type: ignore[attr-defined]
```
jaeopt, line 1384: What are the benefits of waiting for datafile polling thread stop?

jaeopt (changes requested):
Wondering what benefits we'll have with this fix?

optimizely/optimizely.py
andrewleap-optimizely: Prevents it from throwing errors- mostly during testing

andrewleap-optimizely:
> Wondering what benefits we'll have with this fix?

The main benefit of this change is the PollingConfigManager.stop method- not necessarily required to add it to Optimizely.close.

The benefit is that we can remove the hacky wrappers that were necessary to stop the intermittent failing tests and more closely test the real world behavior.

Mat001 (approved):
LGTM

PR #411, opened 2022-11-23
feat: fetch timeout made configurable

ozayr-zaviar:
Summary
-------
- Fetch segment and send opd event timeout made configurable through `sdk_settings` option

Test plan
---------
- All unit tests and FSC should pass.

Ticket
------

[FSSDK-8693](https://jira.sso.episerver.net/browse/FSSDK-8693)

### DIFF
--- a/optimizely/odp/odp_event_api_manager.py
+++ b/optimizely/odp/odp_event_api_manager.py
@@ -43,7 +43,7 @@ class OdpEventApiManager:
     def __init__(self, logger: Optional[optimizely_logger.Logger] = None):
         self.logger = logger or optimizely_logger.NoOpLogger()
 
-    def send_odp_events(self, api_key: str, api_host: str, events: list[OdpEvent]) -> bool:
+    def send_odp_events(self, api_key: str, api_host: str, events: list[OdpEvent], odp_event_timeout: Optional[int] = None) -> bool:
         """
         Dispatch the event being represented by the OdpEvent object.
 
@@ -51,6 +51,7 @@ class OdpEventApiManager:
           api_key: public api key
           api_host: domain url of the host
           events: list of odp events to be sent to optimizely's odp platform.
+          odp_event_timeout: event request timeout in seconds (Optional).
 
         Returns:
             retry is True - if network or server error (5xx), otherwise False
@@ -69,7 +70,7 @@ class OdpEventApiManager:
             response = requests.post(url=url,
                                      headers=request_headers,
                                      data=payload_dict,
-                                     timeout=OdpEventApiConfig.REQUEST_TIMEOUT)
+                                     timeout=odp_event_timeout or OdpEventApiConfig.REQUEST_TIMEOUT)
 
             response.raise_for_status()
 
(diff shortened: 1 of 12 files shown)


andrewleap-optimizely (changes requested):
Looks good! A couple minor suggestions.

Also can we add a test confirming the new settings get passed in appropriately?

optimizely/odp/odp_event_api_manager.py
```diff
@@ -43,14 +43,19 @@ class OdpEventApiManager:
     def __init__(self, logger: Optional[optimizely_logger.Logger] = None):
         self.logger = logger or optimizely_logger.NoOpLogger()
 
-    def send_odp_events(self, api_key: str, api_host: str, events: list[OdpEvent]) -> bool:
+    def send_odp_events(self,
```
jaeopt, line 46: What about passing the timeout value when initializing?  We do not need to send timeout for every event. It's not expected we need event-by-event control. Also a custom OdpEventApiManager can have own timeout control.

optimizely/odp/odp_segment_api_manager.py
```diff
@@ -112,7 +112,8 @@ def __init__(self, logger: Optional[optimizely_logger.Logger] = None):
         self.logger = logger or optimizely_logger.NoOpLogger()
 
     def fetch_segments(self, api_key: str, api_host: str, user_key: str,
-                       user_value: str, segments_to_check: list[str]) -> Optional[list[str]]:
+                       user_value: str, segments_to_check: list[str],
```
jaeopt, line 115: same here. we may keep the timeout constant locally instead of passing timeout for each api call.

optimizely/odp/odp_manager.py
```diff
@@ -56,6 +59,7 @@ def __init__(
             self.segment_manager = OdpSegmentManager(segments_cache, logger=self.logger)
 
         self.event_manager = self.event_manager or OdpEventManager(self.logger)
+        self.event_manager.odp_event_timeout = odp_event_timeout
```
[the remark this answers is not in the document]
jaeopt: +1  Pass it down all the way to OdpEventApiManager and OdpSegmentApiManager?

jaeopt (changes requested):
A suggestion to pass down timeout while initialization. Other than that, it looks good.

optimizely/odp/odp_segment_api_manager.py
```diff
@@ -108,8 +108,9 @@
 class OdpSegmentApiManager:
     """Interface for manging the fetching of audience segments."""
 
-    def __init__(self, logger: Optional[optimizely_logger.Logger] = None):
+    def __init__(self, logger: Optional[optimizely_logger.Logger] = None, timeout: Optional[int] = None):
         self.logger = logger or optimizely_logger.NoOpLogger()
+        self.timeout = timeout
```
andrewleap-optimizely, line 113:
```suggestion
        self.timeout = timeout or OdpSegmentApiConfig.REQUEST_TIMEOUT
```

andrewleap-optimizely (changes requested):
Just a couple more small suggestions. Also can we add one more test for Optimizely initialization with OptimizelySdkSettings and the two new parameters?

jaeopt (approved):
LGTM

andrewleap-optimizely (approved):
lgtm

PR #413, opened 2023-01-19
fix: add notification center registry

andrewleap-optimizely:
Summary
-------
Create a private registry of NotificationCenters for internal use.
- Remove dependence on user supplied NotificationCenter.
- Resolve issue of notifications being registered with one NotificationCenter and sent to a different one.

Test plan
---------
added tests to
- test_notification_center_registry.py
- test_optimizely.py
- test_optimizely_config.py
- test_optimizely_factory.py

Ticket
------

[FSSDK-8773](https://jira.sso.episerver.net/browse/FSSDK-8773)

### DIFF
--- a/optimizely/optimizely.py
+++ b/optimizely/optimizely.py
@@ -164,7 +164,7 @@ class Optimizely:
                 self.config_manager = StaticConfigManager(**config_manager_options)
 
         self.odp_manager: OdpManager
-        self.setup_odp(sdk_key or self.config_manager.get_sdk_key())
+        self.setup_odp(self.config_manager.get_sdk_key())
 
         self.event_builder = event_builder.EventBuilder()
         self.decision_service = decision_service.DecisionService(self.logger, user_profile_service)
--- a/optimizely/optimizely.py
+++ b/optimizely/optimizely.py
@@ -1,4 +1,4 @@
-# Copyright 2016-2022, Optimizely
+# Copyright 2016-2023, Optimizely
 # Licensed under the Apache License, Version 2.0 (the "License");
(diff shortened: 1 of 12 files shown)


optimizely/optimizely.py
```diff
@@ -174,8 +163,8 @@ def __init__(
             else:
                 self.config_manager = StaticConfigManager(**config_manager_options)
 
-        if not self.sdk_settings.odp_disabled:
-            self._update_odp_config_on_datafile_update()
+        self.odp_manager: OdpManager
+        self.setup_odp(sdk_key or self.config_manager.get_sdk_key())
```
jaeopt, line 167: I see 2 changes related to "sdk_key" in here and in config_manager, both optional.
If none of them required, then we may have a case sdk_key is null.
As noted in my comment in config_manager, we may not need get_sdk_key in config_manager. 
We can consider "sdk_key" in optimizely from optional to required while removing the changes in config_manager.

jaeopt (changes requested):
Changes look good.  I see a case when sdk_key is not provided. See my comments.

tests/test_notification_center_registry.py
```diff
@@ -0,0 +1,84 @@
+# Copyright 2019, Optimizely
```
Mat001, line 1: 2023?

Mat001 (approved):
Looks ok. One typo

jaeopt (approved):
LGTM

jaeopt:
@andrewleap-optimizely can we keep a note about this breaking change. We should release a major version bump with this note.

PR #417, opened 2023-02-22
fix: change batch_size from 0 to 1

andrewleap-optimizely:
Summary
-------

-  When flush_interval is zero, set batch_size to 1.

Test plan
---------
All tests pass

Ticket
------

-  [FSSDK-8804](https://jira.sso.episerver.net/browse/FSSDK-8804)


jaeopt (approved):
LGTM

PR #418, opened 2023-02-24
[FSSDK-8946] fix: make odp event identifiers required

andrewleap-optimizely:
Summary
-------
-  Updated send_odp_events to require identifiers and contain at least one key

Test plan
---------
- updated test_optimizely.py

Ticket
------

-  [FSSDK-8946](https://jira.sso.episerver.net/browse/FSSDK-8946)

### DIFF
--- a/optimizely/optimizely.py
+++ b/optimizely/optimizely.py
@@ -1369,8 +1369,8 @@ class Optimizely:
     def send_odp_event(
         self,
         action: str,
+        identifiers: dict[str, str],
         type: str = enums.OdpManagerConfig.EVENT_TYPE,
-        identifiers: Optional[dict[str, str]] = None,
         data: Optional[dict[str, str | int | float | bool | None]] = None
     ) -> None:
         """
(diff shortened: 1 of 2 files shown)


jaeopt (approved):
LGTM!

optimizely/optimizely.py
```diff
@@ -1369,25 +1369,29 @@ def _fetch_qualified_segments(self, user_id: str, options: Optional[list[str]] =
     def send_odp_event(
         self,
         action: str,
+        identifiers: dict[str, str],
```
Mat001, line 1372: Is there a reason we moved identifiers higher in the arg list?
I believe it should be in the third place? I've tested without keyword args where order mattered ...
image

Mat001 (changes requested):
Other than the order, it looks good

optimizely/optimizely.py
andrewleap-optimizely: Same reason why `action` is first instead of `type`- parameters with defaults must come after non-default parameters.

Mat001 (approved):
lgtm

PR #419, opened 2023-03-09
fix: block odp methods on datafile

andrewleap-optimizely:
Summary
-------

- Added blocking call to `Optimizely._identify`, `Optimizely._fetch_qualified_segments` and `Optimizely.send_odp_event` to ensure datafile has been downloaded before attempting to make odp requests.

Test plan
---------
all existing tests pass

### DIFF
--- a/optimizely/optimizely.py
+++ b/optimizely/optimizely.py
@@ -1357,11 +1357,25 @@ class Optimizely:
             self.logger.error(enums.Errors.INVALID_OPTIMIZELY.format('identify_user'))
             return
 
+        if not self.odp_manager.enabled:
+            self.logger.debug('ODP identify event is not dispatched (ODP disabled).')
+            return
+
+        config = self.config_manager.get_config()
+        if not config:
+            self.logger.error(enums.Errors.INVALID_PROJECT_CONFIG.format('identify_user'))
+            return
+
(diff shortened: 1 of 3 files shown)


optimizely/optimizely.py
```diff
@@ -1357,13 +1357,27 @@ def _identify_user(self, user_id: str) -> None:
             self.logger.error(enums.Errors.INVALID_OPTIMIZELY.format('identify_user'))
             return
 
+        if not self.odp_manager.enabled:
+            self.logger.debug('ODP identify event is not dispatched (ODP disabled).')
+            return
```
jaeopt, line 1362: I believe this is checked in odp_manager. Do we need to check here again?

jaeopt (approved):
LGTM - a clarification

Mat001 (approved):
Looks good. Tested fetch without calling opti config.

optimizely/optimizely.py
andrewleap-optimizely: This allows `create_user_context` to remain non-blocking if odp is disabled. Seems like a worthwhile optimization?

jaeopt: @andrewleap-optimizely I see. Probably not in server sdks. They'll be blocked anyway in following decide call even if avoid it here :)

PR #423, opened 2023-04-07
[FSSDK-9069] fix: odp event validation

andrewleap-optimizely:
Summary
-------
Add validation to send_odp_events:

- log error when action == None or ""
- use default when type == None or ""
- fix case and/or "-" for `fs_user_id` identifier key

Test plan
---------
added tests to:
test_odp_event_manager.py
test_optimizely.py

Ticket
------
[FSSDK-9069](https://jira.sso.episerver.net/browse/FSSDK-9069)

### DIFF
--- a/optimizely/helpers/enums.py
+++ b/optimizely/helpers/enums.py
@@ -126,6 +126,7 @@ class Errors:
     ODP_NOT_INTEGRATED: Final = 'ODP is not integrated.'
     ODP_NOT_ENABLED: Final = 'ODP is not enabled.'
     ODP_INVALID_DATA: Final = 'ODP data is not valid.'
+    ODP_INVALID_ACTION: Final = 'ODP action is not valid (cannot be empty).'
     MISSING_SDK_KEY: Final = 'SDK key not provided/cannot be found in the datafile.'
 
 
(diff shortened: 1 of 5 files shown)


jaeopt (approved):
LGTM

Mat001 (approved):
LGTM - one question: we decided to not check for a mix of dash and underscore, correct? i.e. FS-USER_ID
Most common mistake is probably all dashes and all uppercases. The mx of dash/underscore or a mix of upper and lower case would be a diff kind of category, quite unlikely I think
@jaeopt @andrewleap-optimizely

andrewleap-optimizely:
> LGTM - one question: we decided to not check for a mix of dash and underscore, correct? i.e. FS-USER_ID Most common mistake is probably all dashes and all uppercases. The mx of dash/underscore or a mix of upper and lower case would be a diff kind of category, quite unlikely I think @jaeopt @andrewleap-optimizely

Mixed case is covered by this implementation. We could add support for mixed underscore/dash but seems an unlikely scenario in my opinion.

jaeopt:
> > LGTM - one question: we decided to not check for a mix of dash and underscore, correct? i.e. FS-USER_ID Most common mistake is probably all dashes and all uppercases. The mx of dash/underscore or a mix of upper and lower case would be a diff kind of category, quite unlikely I think @jaeopt @andrewleap-optimizely
> 
> Mixed case is covered by this implementation. We could add support for mixed underscore/dash but seems an unlikely scenario in my opinion.

Agreed. Mixed underscore/dash looks like obvious user errors, not from our ambiguity :)

PR #430, opened 2023-12-05
[FSSDK-9780] Return Latest Experiment When Duplicate Keys in Config

Mat001:
Summary
-------

-  As a developer, I want to avoid non-deterministic return values in optimizely_get_config method so that I can anticipate the return values in my application logic.
- The problem: we have flag 1 and flag 2. We have experiment rule A with id 123 and experiment rule A with id 456 (same rule key). Flag 1 has the first experiment rule A, flag 2 has the second experiment rule A. Has a rule with the same experiment key.
We want to avoid that, so we provide a check and a warning log.

Test plan
---------
- write unit tests

Issues
------

- https://jira.sso.episerver.net/browse/FSSDK-9780


jaeopt (approved):
LGTM

optimizely/optimizely_config.py
```diff
@@ -47,6 +50,7 @@ def __init__(
         self.attributes = attributes or []
         self.events = events or []
         self.audiences = audiences or []
+        self.logger = logger
```
andrewleap-optimizely, line 53: Where did this logger come from? (not where it was supposed to!)

andrewleap-optimizely (changes requested):

tests/test_optimizely_config.py
```diff
@@ -1473,6 +1474,81 @@ def test__get_experiments_maps(self):
...
+
+        # add new experiment rule with the same key and a new feature with the same rule key
+        self.config_dict_with_features['experiments'].append(new_experiment)
+        self.config_dict_with_features['featureFlags'].append(new_feature)
+
+        config_with_duplicate_key = self.config_dict_with_features
+        opt_instance = optimizely.Optimizely(json.dumps(config_with_duplicate_key))
+        self.project_config = opt_instance.config_manager.get_config()
+
+        with patch('optimizely.logger.SimpleLogger.warning') as mock_logger:
```
Mat001, line 1530: Ideally it'd be patching smth like: 'optimizely.optimizely_config.OptimizelyConfigService.logger'.
SimpleLogger looks a bit weird. But I'm ok if it's just for the test

tests/test_optimizely_config.py
```diff
@@ -1746,7 +1750,7 @@ def test_get_audiences(self):
             error_handler=None
         )
 
-        config_service = optimizely_config.OptimizelyConfigService(proj_conf)
+        config_service = optimizely_config.OptimizelyConfigService(proj_conf, logger)
```
andrewleap-optimizely, line 1753: This is the logger module ☝️

andrewleap-optimizely (changes requested):

andrewleap-optimizely (approved):
LGTM

PR #433, opened 2024-03-11
Mpirnovar update error

Mat001:
Summary
-------

-  Update error log message, make it more accurate

Test plan
---------

Issues
------

-  [FSSDK-10064](https://jira.sso.episerver.net/browse/FSSDK-10064)

### DIFF
--- a/optimizely/config_manager.py
+++ b/optimizely/config_manager.py
@@ -422,7 +422,7 @@ class PollingConfigManager(StaticConfigManager):
                     break
         except (OSError, OverflowError) as err:
             self.logger.error(
-                f'Provided update_interval value may be too big. Error: {err}'
+                f'Thread for background datafile polling failed. Error: {err}'
             )
             raise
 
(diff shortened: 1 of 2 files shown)


optimizely/config_manager.py
```diff
@@ -422,7 +422,7 @@ def _run(self) -> None:
                     break
         except (OSError, OverflowError) as err:
```
andrewleap-optimizely, line 423: Seems like we should change the exception to be a catch all?

Mat001:
@andrewleap-optimizely (cc @pvcraven ) Any idea why with "self.assertRaises(OverflowError):" would not catch the Overflow error. But logs when Overflow exception occurs do show. It's all in the _run() function in config_manager().

I trigger Overflow error by passing a very large (overflowing) int to the update_interval for polling. 

What confuses me is that the logs from that try block in the _run() funciton do show, but the Error is not raised.
The assumption is that I mock the right thing.

Here I moch fetch_datafile get request that is inside PollingConfigManager:

project_config_manager = config_manager.PollingConfigManager(sdk_key=sdk_key,
                                                                             logger=mock_logger,
                                                                             update_interval=12345678912345)
                                                                             
However the error happens in the _run() function that is part of fetch_datafile(). Maybe error doesn't bubble up?

andrewleap-optimizely (approved):

PR #453, opened 2025-05-06
[FSSDK-11148] update: Implement CMAB Client

FarhanAnjum-opti:
Summary
-------

- Implemented DefaultCmabClient, a client for fetching CMAB (Contextual Multi-Armed Bandit) decisions.

- Supports configurable retries with exponential backoff through CmabRetryConfig.

- Handles error scenarios such as network failures, non-2xx responses, invalid JSON, and malformed responses.

- Logs relevant error and retry information using the Optimizely logger.

Test plan
---------
- Created unit tests for checking various scenarios.

Issues
------

-  [FSSDK-1148](https://jira.sso.episerver.net/browse/FSSDK-11148)

### DIFF
--- a/optimizely/cmab/cmab_client.py
+++ b/optimizely/cmab/cmab_client.py
@@ -73,7 +73,7 @@ class DefaultCmabClient:
         user_id: str,
         attributes: Dict[str, Any],
         cmab_uuid: str,
-        timeout: Optional[float] = None
+        timeout: float = MAX_WAIT_TIME
     ) -> str:
         """Fetch a decision from the CMAB prediction service.
 
@@ -82,13 +82,12 @@ class DefaultCmabClient:
             user_id (str): The user ID for the request.
             attributes (Dict[str, Any]): User attributes for the request.
             cmab_uuid (str): Unique identifier for the CMAB request.
-            timeout (float): Maximum wait time for request to respond in seconds.
+            timeout (float): Maximum wait time for request to respond in seconds. Defaults to 10 seconds.
 
         Returns:
             str: The variation ID.
         """
         url = CMAB_PREDICTION_ENDPOINT % rule_id
-        timeout = timeout or MAX_WAIT_TIME
         cmab_attributes = [
             {"id": key, "value": value, "type": "custom_attribute"}
             for key, value in attributes.items()
--- a/optimizely/cmab/cmab_client.py
+++ b/optimizely/cmab/cmab_client.py
@@ -17,6 +17,7 @@ import math
 from typing import Dict, Any, Optional
 from optimizely import logger as _logging
 from optimizely.helpers.enums import Errors
+from optimizely.exceptions import CmabFetchError, CmabInvalidResponseError
 
 # CMAB_PREDICTION_ENDPOINT is the endpoint for CMAB predictions
 CMAB_PREDICTION_ENDPOINT = "https://prediction.cmab.optimizely.com/predict/%s"
@@ -71,8 +72,9 @@ class DefaultCmabClient:
         rule_id: str,
         user_id: str,
         attributes: Dict[str, Any],
-        cmab_uuid: str
-    ) -> Optional[str]:
+        cmab_uuid: str,
+        timeout: Optional[float] = None
+    ) -> str:
         """Fetch a decision from the CMAB prediction service.
 
         Args:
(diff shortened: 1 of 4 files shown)


optimizely/cmab/cmab_client.py
```diff
@@ -0,0 +1,191 @@
...
+
+
+class DefaultCmabClient:
+    """Client for interacting with the CMAB service.
+
+    Provides methods to fetch decisions with optional retry logic.
+    """
+    def __init__(self, http_client: Optional[requests.Session] = None,
+                 retry_config: Optional[CmabRetryConfig] = None,
+                 logger: Optional[_logging.Logger] = None):
```
Mat001, line 57: I wonder if we add timeout here (from requests library). because js sdk and go-sdk somewhat handle cacellation.
The JavaScript implementation doesn't explicitly handle cancellation like Go does, but it does use Promises which can be cancelled by the caller. Python should probably at least support configurable timeouts? maybe by adding a timeout arg? So if a request takes too long can be properly cancelled and processed as error/exception.

timeout: float = MAX_WAIT_TIME

and then self.timeout = timeout in the assignment.

Mat001: then in fetch_decision function you can do this:
```suggestion
                 
    def fetch_decision(self, rule_id, user_id, attributes, cmab_uuid, timeout=None):
        # Use the timeout parameter or fall back to default
        timeout = timeout or MAX_WAIT_TIME
    
        # Pass timeout to requests
        response = self.http_client.post(url, data=json.dumps(request_body), 
                                        headers=headers, timeout=timeout)
```

tests/test_cmab_client.py
```diff
@@ -0,0 +1,161 @@
...
+    def test_fetch_decision_request_exception(self, mock_do_fetch):
+        result = self.client.fetch_decision(self.rule_id, self.user_id, self.attributes, self.cmab_uuid)
+        self.assertIsNone(result)
+        self.mock_logger.error.assert_called_with(Errors.CMAB_FETCH_FAILED.format("Network error"))
+
+    @patch.object(DefaultCmabClient, '_do_fetch', return_value=None)
+    def test_fetch_decision_invalid_response(self, mock_do_fetch):
+        result = self.client.fetch_decision(self.rule_id, self.user_id, self.attributes, self.cmab_uuid)
+        self.assertIsNone(result)
+        self.mock_logger.error.assert_called_once()
```
Mat001, line 161: in tests I'd add validating raised exceptions that you may add per my comments?.

Mat001 (changes requested):

optimizely/cmab/cmab_client.py
```diff
@@ -0,0 +1,203 @@
...
+            rule_id (str): The rule ID for the experiment.
+            user_id (str): The user ID for the request.
+            attributes (Dict[str, Any]): User attributes for the request.
+            cmab_uuid (str): Unique identifier for the CMAB request.
+            timeout (float): Maximum wait time for request to respond in seconds. Defaults to 10 seconds.
+
+        Returns:
+            str: The variation ID.
+        """
+        url = CMAB_PREDICTION_ENDPOINT % rule_id
```
Mat001, line 90: Maybe use f-strng here to be a little more readable?
rl = f"/api/cmab/rules/{rule_id}/prediction"

Mat001 (approved):
looks good. Just one comment about string format

optimizely/cmab/cmab_client.py
```diff
@@ -0,0 +1,200 @@
...
+        }
+
+        try:
+            if self.retry_config:
+                variation_id = self._do_fetch_with_retry(url, request_body, self.retry_config, timeout)
+            else:
+                variation_id = self._do_fetch(url, request_body, timeout)
+            return variation_id
+
+        except requests.RequestException as e:
```
raju-opti, line 109: _do_fetch and _do_fetch_with_retry already handles the RequestException. Those methods should handle the RequestException and wrap it in custom exception and raise. We can remove try .. except from this method.

tests/test_cmab_client.py
```diff
@@ -0,0 +1,169 @@
...
+            http_client=self.mock_http_client,
+            logger=self.mock_logger,
+            retry_config=None
+        )
+        self.rule_id = 'test_rule'
+        self.user_id = 'user123'
+        self.attributes = {'attr1': 'value1'}
+        self.cmab_uuid = 'uuid-1234'
+
+    @patch.object(DefaultCmabClient, '_do_fetch', return_value='var-abc')
```
raju-opti, line 148: Can we not mock internal details (_do_fetch and _do_fetch_with_retry) and instead use the mock_http_client to control the api return value?

tests/test_cmab_client.py
```diff
@@ -0,0 +1,169 @@
...
+    def test_fetch_decision_success_no_retry(self, mock_do_fetch):
+        result = self.client.fetch_decision(self.rule_id, self.user_id, self.attributes, self.cmab_uuid)
+        self.assertEqual(result, 'var-abc')
+        mock_do_fetch.assert_called_once()
```
raju-opti, line 152: we also need to check that the correct parameters like the correct userId, experimentId, attributes, cmabUUID are passed to the api

raju-opti (changes requested):

raju-opti (approved):

PR #459, opened 2025-07-02
[FSSDK-11458] Python - Add SDK Multi-Region Support for Data Hosting

esrakartalOpt:
Summary
-------

-  Add Multi-Region Support for Data Hosting

Test plan
---------

- Created and added new test cases

Issues
------

- [FSSDK-11458](https://jira.sso.episerver.net/browse/FSSDK-11458)

### DIFF
--- a/optimizely/project_config.py
+++ b/optimizely/project_config.py
@@ -41,6 +41,7 @@ RESERVED_ATTRIBUTE_PREFIX: Final = '$opt_'
 
 EntityClass = TypeVar('EntityClass')
 
+
 class ProjectConfig:
     """ Representation of the Optimizely project config. """
 


tests/test_user_event_factory.py
```diff
@@ -41,6 +41,21 @@ def test_impression_event(self):
...
+        user_id = 'test_user'
+
+        project_config.region = 'EU'
+
+        impression_event = UserEventFactory.create_impression_event(
+            project_config, experiment, '111128', '', 'rule_key', 'rule_type', True, user_id, None
+        )
+
+        self.assertEqual(self.project_config.region, impression_event.event_context.region)
+        self.assertEqual('EU', impression_event.event_context.region)
```
raju-opti, line 58: can we also add a test for create_contervsion_event?

tests/test_event_processor.py
```diff
@@ -576,7 +576,7 @@ def __init__(self, is_updated=False):
         self.is_updated = is_updated
 
     def dispatch_event(self, log_event):
-        if log_event.http_verb == 'POST' and log_event.url == EventFactory.EVENT_ENDPOINT:
+        if log_event.url in EventFactory.EVENT_ENDPOINTS.values():
```
raju-opti, line 579: the log_event.http_verb == 'POST' condition got removed. Can we keep that check?

tests/test_event_factory.py
```diff
@@ -918,5 +966,9 @@ def test_create_conversion_event__when_event_is_used_in_multiple_experiments(sel
         log_event = EventFactory.create_log_event(event_obj, self.logger)
 
         self._validate_event_object(
-            log_event, EventFactory.EVENT_ENDPOINT, expected_params, EventFactory.HTTP_VERB, EventFactory.HTTP_HEADERS,
+            log_event,
+            EventFactory.EVENT_ENDPOINTS.get('US'),
+            expected_params,
+            EventFactory.HTTP_VERB,
+            EventFactory.HTTP_HEADERS,
         )
```
raju-opti, line 974: can we add a test to check that create_log_event works correctly with EU region?

raju-opti (changes requested):
few small comments

optimizely/event/event_factory.py
```diff
@@ -97,7 +100,18 @@ def create_log_event(
...
         event_params = event_batch.get_event_params()
 
-        return log_event.LogEvent(cls.EVENT_ENDPOINT, event_params, cls.HTTP_VERB, cls.HTTP_HEADERS)
+        region = user_context.region
+        if hasattr(region, 'value'):
+            region_str = region.value
+        elif region is None:
+            region_str = 'US'  # Default to US
+        else:
+            region_str = str(region)
```
jaeopt, line 109: We can consider using a simple string type for region. The region will be handled in type-safe way when building into datafile. It looks like fallback to "US" for safety is good enough?

optimizely/project_config.py
```diff
@@ -85,6 +91,13 @@ def __init__(self, datafile: str | bytes, logger: Logger, error_handler: Any):
         self.host_for_odp: Optional[str] = None
         self.all_segments: list[str] = []
 
+        region_value = config.get('region')
+        self.region: Region
+        if region_value == Region.EU.value:
+            self.region = Region.EU
+        else:
+            self.region = Region.US
```
jaeopt, line 99: Another area we can get benefits with string type. No need to change the code when we add countries.

jaeopt (changes requested):
Looks good! We can consider string-type region for simple.

optimizely/event/user_event.py
```diff
@@ -99,10 +100,11 @@ def __init__(
...
 
-    def __init__(self, account_id: str, project_id: str, revision: str, anonymize_ip: bool):
+    def __init__(self, account_id: str, project_id: str, revision: str, anonymize_ip: bool, region: Region):
         self.account_id = account_id
         self.project_id = project_id
         self.revision = revision
         self.client_name = CLIENT_NAME
         self.client_version = version.__version__
         self.anonymize_ip = anonymize_ip
+        self.region = region or 'US'
```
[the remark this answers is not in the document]
Mat001: I second htis use. If yuo have Region class somewhere, then we should call country attribute on the Region object.

optimizely/event_builder.py
```diff
@@ -266,7 +269,10 @@ def create_impression_event(
 
         params[self.EventParams.USERS][0][self.EventParams.SNAPSHOTS].append(impression_params)
 
-        return Event(self.EVENTS_URL, params, http_verb=self.HTTP_VERB, headers=self.HTTP_HEADERS)
+        region = project_config.region.value if hasattr(project_config.region, 'value') else 'US'
+        events_url = self.EVENTS_URLS.get(str(region), self.EVENTS_URLS['US'])
```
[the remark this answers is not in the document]
Mat001: Again, I agree with copilot lol

optimizely/event/event_factory.py
Mat001: @jaeopt by string type do you mean using Region.US instead of string 'US'?
That's what I have been suggesting further down the code. Just we don't contradict each other.

PR #505, opened 2026-04-16
[FSSDK-12497] Fix return in finally block silently swallowing exceptions

### LINKED ISSUE #439: return in finally can swallow exception
In https://github.com/optimizely/python-sdk/blob/40880ffad7403ef96c7b11b02a110fb42adf39c2/optimizely/config_manager.py#L153  there is a `return` statement in a `finally` block, which would swallow any in-flight exception.

This means that if any exception is raised from one of the `except:` clauses, it will not propagate on as expected.

See also https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions.

FarhanAnjum-opti:
## Summary

Fixes a bug where a `return` statement inside a `finally` block in `_set_config` silently swallows exceptions. Per [Python docs](https://docs.python.org/3/tutorial/errors.html#defining-clean-up-actions), a `return` in `finally` suppresses any in-flight exception, meaning errors raised inside `except` blocks are completely lost with no traceback.

## Changes

- Moved error-handling logic out of `finally` into normal control flow after the try/except block
- Changed bare `except:` to `except Exception:` to avoid catching `KeyboardInterrupt` and `SystemExit`
- Added unit test verifying that exceptions in the `except` block propagate correctly

## Jira Ticket

[FSSDK-12497](https://optimizely-ext.atlassian.net/browse/FSSDK-12497)

## Issues

- Fixes https://github.com/optimizely/python-sdk/issues/439

[FSSDK-12497]: https://optimizely-ext.atlassian.net/browse/FSSDK-12497?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/optimizely/config_manager.py
+++ b/optimizely/config_manager.py
@@ -145,14 +145,14 @@ class StaticConfigManager(BaseConfigManager):
         except optimizely_exceptions.UnsupportedDatafileVersionException as error:
             error_msg = error.args[0]
             error_to_handle = error
-        except:
+        except Exception:
             error_msg = enums.Errors.INVALID_INPUT.format('datafile')
             error_to_handle = optimizely_exceptions.InvalidInputException(error_msg)
-        finally:
-            if error_msg or config is None:
-                self.logger.error(error_msg)
-                self.error_handler.handle_error(error_to_handle or Exception('Unknown Error'))
-                return
+
+        if error_msg or config is None:
+            self.logger.error(error_msg)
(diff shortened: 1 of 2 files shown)


Mat001 (approved):
yeah, looks good

PR #513, opened 2026-05-18
[AI-FSSDK] [FSSDK-12670] Block ODP identify event for single identifier

jaeopt:
## Summary

Fixes ODP identify event to only send when multiple identifiers exist. Previously, the SDK always sent an identify event when creating a user context, even with just a single `fs_user_id`. This wastes network calls and pollutes ODP identity graphs with single-identifier records that provide no linking value.

## Changes
- Changed `identify_user` method signature across the identify chain to accept an identifiers dict instead of a single user_id string
- Added identifier count check in `OdpEventManager.identify_user` to skip sending when fewer than 2 valid identifiers are present
- Filters out identifiers with None or empty string values before counting
- Added debug-level logging when the identify event is skipped due to single identifier
- Updated existing tests and added new tests for single-identifier blocking and empty-value filtering

## Jira Ticket
[FSSDK-12670](https://optimizely-ext.atlassian.net/browse/FSSDK-12670)

[FSSDK-12670]: https://optimizely-ext.atlassian.net/browse/FSSDK-12670?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/optimizely/odp/odp_event_manager.py
+++ b/optimizely/odp/odp_event_manager.py
@@ -267,9 +267,28 @@ class OdpEventManager:
         except Full:
             self.logger.warning(Errors.ODP_EVENT_FAILED.format("Queue is full"))
 
-    def identify_user(self, user_id: str) -> None:
+    def identify_user(self, identifiers: dict[str, str]) -> None:
+        """Send an identify event to ODP if there are multiple valid identifiers.
+
+        An identify event is only useful when there are two or more identifiers
+        to link together in the ODP user profile. Sending a single identifier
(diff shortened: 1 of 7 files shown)


FarhanAnjum-opti (approved):
LGTM. Confirmed the < 2 guard intentionally blocks the single fs_user_id from create_user_context. The dict refactoring is prep for future multi-identifier support.

PR #517, opened 2026-06-12
[AI-FSSDK] [FSSDK-12760] Add localHoldouts datafile section for backward compat

jaeopt:
## Summary

Adds support for the new top-level `localHoldouts` datafile section so a single datafile can serve every SDK generation safely. Older SDKs ignore the unknown key; this SDK now parses it as the source of truth for rule-scoped (local) holdouts, while `holdouts` is treated as global-only — section membership is the sole signal for scope.

## Changes

- Parse `localHoldouts` as a separate top-level array; entries register against each rule in `includedRules`.
- Treat every entry in `holdouts` as global; any `includedRules` field on those entries is stripped and ignored.
- Reject invalid local-holdout entries (missing or null `includedRules`) with an error log instead of silently promoting them to global.
- Preserve backward compatibility: datafiles without the `localHoldouts` key continue to behave exactly as before.

## Jira Ticket

[FSSDK-12760](https://optimizely-ext.atlassian.net/browse/FSSDK-12760)

[FSSDK-12760]: https://optimizely-ext.atlassian.net/browse/FSSDK-12760?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/optimizely/entities.py
+++ b/optimizely/entities.py
@@ -233,7 +233,11 @@ class Holdout(BaseEntity):
         self.trafficAllocation = trafficAllocation
         self.audienceIds = audienceIds
         self.audienceConditions = audienceConditions
-        # None = global holdout (applies to all rules); list of rule IDs = local holdout
+        # Per-rule targeting list for local holdouts. Scope is NOT determined by this
+        # field — it is determined by which datafile section the entity was parsed
+        # from ('holdouts' = global, 'localHoldouts' = local). ProjectConfig strips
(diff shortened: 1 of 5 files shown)


FarhanAnjum-opti (approved):

FarhanAnjum-opti (approved):
LGTM

FarhanAnjum-opti (approved):

PR #518, opened 2026-06-25
[AI-FSSDK] [FSSDK-12813] Normalize decision event campaign_id, variation_id, and entity_id

jaeopt:
## Summary

Normalize three decision-event ID fields uniformly across every decision type (experiment, feature test, rollout, holdout) in the event-builder layer so the wire payload is byte-equivalent across SDKs for the same input. `decisions[].campaign_id` and impression `events[].entity_id` fall back to `experiment_id` only when the value is empty string or `None` — any non-empty string (numeric like `"12345"` or opaque like `"default-12345"`, `"layer_abc"`) passes through unchanged. `decisions[].variation_id` retains the strict numeric-string contract and falls back to `None` for empty/whitespace/non-numeric input. The path never logs, raises, or drops dispatch.

## Changes

- Added `optimizely/event/event_id_normalizer.py` with pure helpers (`is_non_empty_string`, `is_numeric_id_string`, `normalize_campaign_id`, `normalize_variation_id`). `normalize_campaign_id` accepts any non-empty string and falls back to `experiment_id` only on empty/`None`; `normalize_variation_id` keeps the strict decimal-digit contract and returns `None` otherwise.
- Wired the normalizer into `event_factory.py` (current) and `event_builder.py` (legacy) on the single impression path so both producers emit identical output.
- Reused the normalized `campaign_id` for `events[].entity_id` so the two fields are byte-equivalent by construction. Typed `Decision.variation_id` as `Optional[str]` so invalid inputs serialize as JSON `null`.
- Audited existing event-builder tests per FR-011 — happy-path fixtures already use numeric IDs; integration tests updated to assert opaque-string passthrough for `campaign_id` / `entity_id` and null normalization for `variation_id`.

## Jira Ticket

[FSSDK-12813](https://optimizely-ext.atlassian.net/browse/FSSDK-12813)

[FSSDK-12813]: https://optimizely-ext.atlassian.net/browse/FSSDK-12813?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/optimizely/event/event_factory.py
+++ b/optimizely/event/event_factory.py
@@ -135,12 +135,6 @@ class EventFactory:
                 if isinstance(event.experiment, entities.Experiment):
                     experiment_layerId = event.experiment.layerId
 
-            # FSSDK-12813: Normalize decision-event IDs uniformly across all
-            # decision types (experiment, feature test, rollout, holdout).
-            # campaign_id falls back to experiment_id when invalid.
-            # variation_id becomes None when invalid.
-            # entity_id on the impression event mirrors campaign_id (FR-009)
-            # so the two fields are byte-equivalent for the same event.
             normalized_campaign_id = event_id_normalizer.normalize_campaign_id(
                 experiment_layerId, experiment_id
             )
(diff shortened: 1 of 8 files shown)


FarhanAnjum-opti (approved):
LGTM

PR #519, opened 2026-06-27
[AI-FSSDK] [FSSDK-12839] Normalize impression event campaign_id, variation_id, entity_id

jaeopt:
## Summary
Impression-event payloads emitted by the Python SDK can carry empty, null, or non-numeric values for `decisions[].campaign_id`, `decisions[].variation_id`, and `events[].entity_id` — most visibly on holdout decisions, which legitimately lack a `layerId`. Downstream event processing requires these fields to be numeric strings (or explicit `null` for `variation_id`), so malformed values are silently dropped or misattributed. This change normalizes those fields at the event-builder layer so every impression event conforms to the wire contract.

## Changes
- Added a `is_numeric_string_id` helper that recognises a non-empty ASCII decimal-digit string as the only valid form for these IDs.
- In both the legacy event builder and the modern event factory, normalize `decisions[].campaign_id` to fall back to the decision's `experiment_id` when the source value is invalid, and normalize `decisions[].variation_id` to `None` when invalid.
- Derive `events[].entity_id` from the same normalized value as `campaign_id` so the two fields are guaranteed byte-equal on every impression payload.
- Limited the change to impression events; conversion events continue to derive `entity_id` from the conversion event's own `id`.

## Jira Ticket
[FSSDK-12839](https://optimizely-ext.atlassian.net/browse/FSSDK-12839)

[FSSDK-12839]: https://optimizely-ext.atlassian.net/browse/FSSDK-12839?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/optimizely/event/event_factory.py
+++ b/optimizely/event/event_factory.py
@@ -1,4 +1,4 @@
-# Copyright 2019, 2022, Optimizely
+# Copyright 2019, 2022, 2026, Optimizely
 # Licensed under the Apache License, Version 2.0 (the "License");
 # you may not use this file except in compliance with the License.
 # You may obtain a copy of the License at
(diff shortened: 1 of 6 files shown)


jaeopt:
redundant

PR #520, opened 2026-06-30
[AI-FSSDK] [FSSDK-12750] Use attribute id instead of key for CMAB prediction requests

raju-opti:
## Summary

The CMAB prediction service was incorrectly using attribute keys instead of attribute IDs when building the filtered attributes map for prediction requests. The prediction endpoint expects the attribute ID in the payload, but the code was sending the attribute key. This fix ensures the correct attribute ID is used as the dictionary key in the filtered attributes map.

## Changes

- Updated the attribute filtering logic in the CMAB service to use attribute ID instead of attribute key as the dictionary key
- Updated relevant test expectations to reflect the corrected behavior

## Jira Ticket
[FSSDK-12750](https://optimizely-ext.atlassian.net/browse/FSSDK-12750)

[FSSDK-12750]: https://optimizely-ext.atlassian.net/browse/FSSDK-12750?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/optimizely/cmab/cmab_service.py
+++ b/optimizely/cmab/cmab_service.py
@@ -163,7 +163,7 @@ class DefaultCmabService:
         for attribute_id in cmab_attribute_ids:
             attribute = project_config.attribute_id_map.get(attribute_id)
             if attribute and attribute.key in user_attributes:
-                filtered_user_attributes[attribute.key] = user_attributes[attribute.key]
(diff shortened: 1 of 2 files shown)


FarhanAnjum-opti (approved):
Looks good

